PR #30, opened 2018-07-24
Add Airflow custom versions guide

tedmiston:
Closes https://github.com/astronomerio/airflow-guides/issues/22.

### DIFF
--- a/guides/airflow-versions.md
+++ b/guides/airflow-versions.md
@@ -1,0 +1,68 @@
+---
+title: "Custom Airflow versions on Astronomer"
+description: "How to run custom Airflow versions on the Astronomer Platform"
+date: 2018-07-25T00:00:00.000Z
+slug: "airflow-versions"
+heroImagePath: "https://cdn.astronomer.io/website/img/guides/TheAirflowUI_preview.png"
+tags: ["Airflow", "Astronomer Platform"]
+---
+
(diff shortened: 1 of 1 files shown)


vparekh94:
@tedmiston a few customers have asked about this - how far do you think it is from being mergeable?

schnie:
This was good in theory but doesn't take into account dependencies and stuff like that. I'm working on an airflow 1.10 image over the next day or two that we can start testing with and start officially supporting.

tedmiston:
@vparekh94 IMO it could be merged now.  I agree with Greg that we could make it easier to use by having a concrete example and iterate instructions around the dependencies and anything that breaks when crossing Airflow versions.

That said, I think an intermediate-advanced user gets the 80/20 on how to run 1.10 on EE with this guide as it's written today.  I wouldn't mind putting this MVP of the guide out in the wild and iterating on it personally.

tedmiston:
@vparekh94 @schnie Can we merge this and then iterate on it for the next release once we launch the new official multi-Airflow version support?  IMO this guide is independent of that since people will still need it to run a custom unsupported version like 1.8, master branch, future RCs, etc.

schnie:
@tedmiston you should add a section about resolving dependencies. So far 1.8, 1.9, and 1.10 all have additional os level packages that need to be installed since things change, and the setup.py groupings change.

PR #33, opened 2018-07-26
Add tag checkout before install for stability

tedmiston:


schnie:
Should we template the version in somehow or just indicate that it's a variable, rather than just putting 0.3.1 in there? Don't want to have to update this every release, ya know?

tedmiston:
Template would be great if possible.  I don't know if we have a feature like includes in this markdown to html generator.

I could also change it to just say `<latest tag>`?

schnie:
Yea `<latest tag>` should be good enough.

tedmiston:
@schnie Updated - feel free to merge if it looks good.

PR #39, opened 2018-08-21
opening pr for scripts bash operator guide

rhwang10:


vparekh94:
@rhwang10 pretty solid, just a few things I think you should add:

1) Links to the BashOperator code that shows it will by default, look in the temp directory. 
2) Link to the line that shows how we copy the entire directory into the docker file:
https://github.com/astronomerio/astronomer/blob/master/docker/platform/airflow/onbuild/Dockerfile
3) Clarify that the `RUN` command will run each time the container builds and each time you deploy.

PR #54, opened 2018-10-15
Update documentation

license2e:


petedejoy:
Thanks for this, @license2e! Going to have some folks from our side give a quick review then we'll merge.

PR #129, opened 2019-11-16
Ported over email notifications doc

vparekh94:
Ports over spacecamp doc with examples of adding email alerts

### DIFF
--- a/guides/error-notifications-in-airflow.md
+++ b/guides/error-notifications-in-airflow.md
@@ -1,0 +1,249 @@
+---
+title: "Error Notifications in Airflow"
+description: "Managing Error Notifications"
+date: 2018-05-21T00:00:00.000Z
+slug: "error-notifications-in-airflow"
+heroImagePath: null
+tags: ["Building DAGs", "Redshift", "IMAP"]
+---
+
+# Error Reporting on Airflow
+
+Email notifications are great for monitoring Airflow workflows. They can be sent for failures, successes, and retries.
+
+
+## Setting Notifications at the DAG level
+Notifications set at the DAG level filter down to each task in the DAG - generally in the `default_args`.
+
+By default, `email_on_failure` is set to `True`
+
+
+```python
+from datetime import datetime
+from airflow import DAG
+
+default_args = {
+    'owner': 'airflow',
+    'start_date': datetime(2018, 1, 30),
+    'email': ['viraj@astronomer.io']
+}
+
+with DAG('sample_dag',
+          default_args=default_args,
+          schedule_interval='@daily',
+          catchup=False) as dag:
+    ...
+```
+
+Any task in this DAG's context will throw a failure email to all addresses in the emails array
+
+## Different Levels of Notifications
+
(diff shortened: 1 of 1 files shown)


guides/error-notifications-in-airflow.md
```diff
@@ -0,0 +1,249 @@
...
+}
+
+with DAG('sample_dag',
+          default_args=default_args,
+          schedule_interval='@daily',
+          catchup=False) as dag:
+    ...
+```
+
+Any task in this DAG's context will throw a failure email to all addresses in the emails array
```
Jake1021: throw -> send

guides/error-notifications-in-airflow.md
```diff
@@ -0,0 +1,249 @@
...
+
+## Different Levels of Notifications
+
+Failure notifications are the most common, but different levels can be set where appropriate.
```
Jake1021: can be set where appropriate -> can be set appropriately

guides/error-notifications-in-airflow.md
```diff
@@ -0,0 +1,249 @@
...
+
+Emails on retries are great for testing if failures are by caused extraneous factors like load on an external system. If this is the case, consider setting `retry_exponential_backoff` to `True`.
```
Jake1021: are by caused -> are caused by
factors like load on -> factors, like load, on

guides/error-notifications-in-airflow.md
```diff
@@ -0,0 +1,249 @@
...
+with DAG('sample_dag',
+          default_args=default_args,
+          schedule_interval='@daily',
+          catchup=False) as dag:
+    ...
+```
+
+## Isolating Tasks
+
+For some use cases, it might be helpful to only have failure emails for certain tasks. The BaseOperator that all Airflow Operators inherit from has support for these arguments if you don't want them defined at the DAG level.
```
Jake1021: inherit from has -> inherit has
Not positive if inherit is the correct word. If it is, remover "from"

guides/error-notifications-in-airflow.md
```diff
@@ -0,0 +1,249 @@
...
+    email_title = "Airflow Task {tak_id} Failed".format(context['task_instance'].task_id)
+
+    email_body = "{task_id} in {dag_id} failed.".format(context['task_instance'].task_id, context['task_instance'].dag_id)
+
+    send_email('you_email@address.com', email_title, email_body)
+```
+
+## Setting Up Alerts in Slack
+
+At Astronomer, we drop Airflow notifications in shared slack channels instead of emails. We find that public channels that are available to everyone add clarity and help crowd-source solutions. There are a few ways to accomplish this:
```
Jake1021: public channels that are available to everyone add -> public channels add
Redundant!

guides/error-notifications-in-airflow.md
vparekh94: Inherit from makes sense here in the context of how python handles classes

PR #130, opened 2019-11-19
Simplified BranchOperator example and changed image link

vparekh94:


gabrieleckers:
I think it would be a good idea to include a comment in the branching section about not having any empty paths.  The Airflow docs call this out and I've seen a few customers run into issues because of it.

> If you want to skip some tasks, keep in mind that you can’t have an empty path, if so make a dummy task.
>
> image
>
> like this, the dummy task “branch_false” is skipped
> 
> image
>
> Not like this, where the join task is skipped
> 
>

vparekh94:
thanks @gabrieleckers ! just made the change

guides/airflow-branch-operator.md
```diff
@@ -7,94 +7,130 @@ heroImagePath: null
...
+            m = DummyOperator(task_id='branch_{0}_{1}'.format(i, j))
+
+            d >> m
+
+        branching >> d
+```
+![skipped](https://assets.astronomer.io/website/img/guides/branching.png)
+The DAG will proceed based on the output of the function passed in. 
+
+**Note:** You **cannot** have an empty path when skipping tasks - the `skipped` state will apply to all tasks donwstream of whatever task is skipped. Depending on your use case, it may make sense to add a DummyOperator downstream of a task tha can be skipped before the branches from the `BranchPythonOperator` meet.
```
gabrieleckers: **Note:** You **cannot** have an empty path when skipping tasks - the `skipped` state will apply to all tasks immediately donwstream of whatever task is skipped. Depending on your use case, it may make sense to add a DummyOperator downstream of a task tha can be skipped before the branches from the `BranchPythonOperator` meet.

vparekh94: fixed! thanks for attention to detail with the wording

PR #131, opened 2019-11-19
Connections snippet

vparekh94:
Added @andrewhharmon 's snippet sent to DN (needs to be verified to work) and an example of calling connections in a task


gabrieleckers:
Overall this looks great, I would personally change the line

> The `conn_id` is **not** a primary key

To read 

> The `conn_id` **does not** have a uniqueness constraint within Airflow

I just think that communicates what you are after a little more clearly?

vparekh94:
Thanks @gabrieleckers ! Good call! I moved that bit of information to the end of the doc !

PR #147, opened 2020-11-16
fix tag capitalization

ryanahamilton:
This single outlier is causing a duplicate on the website tag sidebar. We should probably make the tags case-agnostic on the website too.

image


petedejoy (approved):
LGTM, thanks @ryanahamilton!

PR #151, opened 2020-11-20
Capitalize Python in Talend guide

kentdanas:


guides/airflow-talend-integration.md
```diff
@@ -75,7 +75,7 @@ The host name should be the Talend Cloud API URL. This can vary depending on whi
 
 Next we can create our DAG.
 
-```python
+```Python
```
petedejoy, line 78: Sorry to confuse, but I believe the code block language specification needs to be lower case for our syntax highlighter to pick it up.

petedejoy (changes requested):

guides/airflow-talend-integration.md
kentdanas: whoops, you're right, sorry about that. should be fixed now

petedejoy (approved):

PR #161, opened 2020-12-10
Azure Data Factory guide

kentdanas:
add guide for airflow-azure-data-factory integration


guides/airflow-azure-data-factory-integration.md
```diff
@@ -0,0 +1,171 @@
...
+## How to Execute ADF Pipelines with Airflow
+
+Operators and hooks are the main building blocks of Airflow, and both can be used easily to execute, or interact with, an ADF pipeline.
+
+### Hooks
+
+We recommend using Airflow hooks when interacting with any external system. Hooks are used as a way to abstract the methods you would use against a source system. Airflow does not currently have built-in hooks for ADF, but they have been developed and can be found on [Github](https://github.com/flvndh/airflow/blob/issue/10995/azure-data-factory/airflow/providers/microsoft/azure/hooks/azure_data_factory.py). 
+
+
+It is likely that these will be merged into the Airflow project soon, but in the meantime, you can always  import them separately, which is what we do in the example below.
```
sunkickr, line 34: there is an extra space between "always' and "import" on this line

guides/airflow-azure-data-factory-integration.md
```diff
@@ -0,0 +1,171 @@
...
+            task_id='run_pipeline',
+            python_callable=run_adf_pipeline,
+            op_kwargs={'pipeline_name': 'pipeline1', 'date': yesterday_date}
+         )
+```
+
+There are a few important things to note about this DAG: 
+
+- We have used the ADF hook code from the Github link above and brought it locally, imported as `AzureDataFactoryHook`. Your import path may vary.
+- The `run_pipeline` task uses the PythonOperator that looks for the `pipeline_name` parameter and a date. If you have other pipelines you want to execute in the same DAG, you can add additional tasks in the same manner.
```
sunkickr, line 152: you may need to ad \ in front of "_" to prevent gatsby from rendering text as italic. I am not sure in this case since it is in ` `.  We may just need to publish and see.

guides/airflow-azure-data-factory-integration.md
```diff
@@ -0,0 +1,171 @@
...
+- This DAG requires an Airflow connection (`azure_data_factory_conn`) to connect to your Azure instance and ADF factory. The connection requires your Tenant ID, Subscription ID, Resource Group, Factory, Client ID, and Client secret. They should be entered into the connection like this:
+
+![Airflow ADF Connection](https://assets2.astronomer.io/main/guides/azure-data-factory/adf_airflow_connection.png)
```
sunkickr, line 155: Same issue here with "_" . I've seen gatsby mess up links that have underscores(by italicizing all the text) in them, but I don't know what will happen with photos.

guides/airflow-azure-data-factory-integration.md
kentdanas: Just looked at the SQL guide and it looks like underscores in `` are rendered okay (no italics)

guides/airflow-azure-data-factory-integration.md
kentdanas: Good call on pointing this out. I checked some of the other guides that have images with underscores in the filepath and they seem okay.

petedejoy (approved):

PR #177, opened 2020-12-21
Azure Container Instances + Airflow guide

kentdanas:
Added guide on using the AzureContainerInstances Operator to do container stuff on Azure with Airflow. Adding @jwitz for tech writing review.

### DIFF
--- a/guides/airflow-azure-container-instances.md
+++ b/guides/airflow-azure-container-instances.md
@@ -41,7 +41,7 @@ With these points in mind, we recommend using ACI with the AzureContainerInstanc
 
 ## Example
 
-Using Airflow to create and run an Azure Container Instance is straightforward: You first identify the Azure resource group you want to create the ACI in (or create a new one), then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
+Using Airflow to create and run an Azure Container Instance is straightforward: You first identify the Azure resource group you want to create the Azure Container Instance in (or create a new one), then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
 
 Next, create an Airflow connection with the type `Azure Container Instance`. Specify your Client ID in the Login field, Client Secret in the Password field, and Tenant and Subscription IDs in the Extras field as json. It should look something like this:
 
--- a/guides/airflow-azure-container-instances.md
+++ b/guides/airflow-azure-container-instances.md
@@ -20,8 +20,8 @@ The easiest way to orchestrate Azure Container Instances with Airflow is to use 
 
 The only prerequisites for using this operator are:
 
-- An Azure account with a resource group created
-- An Azure Service Principle that has write permissions over that resource group
+- An Azure account with a [resource group](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/manage-resource-groups-portal) created
+- An [Azure Service Principle](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals) that has write permissions over that resource group
 - A docker image to use for the container (either publicly or privately available)
 
 This operator can also be used to run existing container instances and make certain updates, including the docker image, environment variables, or commands. Some updates to existing container groups are not possible with the operator, including CPU, memory, and GPU; those updates require deleting the existing container group and recreating it, which can be accomplished using the [AzureContainerInstanceHook](https://github.com/apache/airflow/blob/master/airflow/providers/microsoft/azure/hooks/azure_container_instance.py).
@@ -41,7 +41,7 @@ With these points in mind, we recommend using ACI with the AzureContainerInstanc
 
 ## Example
 
-Using Airflow to create and run an ACI is straightforward: You first identify the Azure resource group you want to create the ACI in (or create a new one), then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
+Using Airflow to create and run an Azure Container Instance is straightforward: You first identify the Azure resource group you want to create the ACI in (or create a new one), then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
 
 Next, create an Airflow connection with the type `Azure Container Instance`. Specify your Client ID in the Login field, Client Secret in the Password field, and Tenant and Subscription IDs in the Extras field as json. It should look something like this:
 
--- a/guides/airflow-azure-container-instances.md
+++ b/guides/airflow-azure-container-instances.md
@@ -43,6 +43,8 @@ With these points in mind, we recommend using ACI with the AzureContainerInstanc
 
 Using Airflow to create and run an Azure Container Instance is straightforward: You first identify the Azure resource group you want to create the Azure Container Instance in (or create a new one), then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
 
+> Note: In Airflow 2.0, provider packages are separate from the core of Airflow. If you are running 2.0 with Astronomer, the `apache-airflow-providers-microsoft-azure` package is already included in our Airlfow Certified Image; if you are not using Astronomer you may need to install this package separately to use the hooks, operators, and connections described here.
+
 Next, create an Airflow connection with the type `Azure Container Instance`. Specify your Client ID in the Login field, Client Secret in the Password field, and Tenant and Subscription IDs in the Extras field as json. It should look something like this:
 
 ![ACI Connection](https://assets2.astronomer.io/main/guides/azure-container-instances/aci_connection.png)
--- a/guides/airflow-azure-container-instances.md
+++ b/guides/airflow-azure-container-instances.md
@@ -28,17 +28,12 @@ Note that this operator can also be used to run existing container instances and
 
 ### When to use ACI
 
-There are multiple ways to work with containers with Airflow on Azure; which you choose depends on your use case. 
-
-The most flexible and scalable method would be to use the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow/stable/kubernetes.html). This method:
-
-- Lets you run any container with Airflow as a Kubernetes pod, which allows you to pass in resource requests and other Kubernetes native parameters
-- Allows you to run containers in different regions
-- Requires an [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/) cluster
+There are multiple ways to work with containers with Airflow on Azure, each of which could be the right fit from your use case. The most flexible and scalable method would be to the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow/stable/kubernetes.html). This will not only let you run any container with Airflow, but also run it as a Kubernetes pod, which allows you to pass in resource requests and other Kubernetes native parameters. However, it requires an [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/) cluster (or a hand rolled Kubernetes cluster).
 
 If you are not running on [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/), ACI can be a great choice:
 
 - It requires very little setup, and is easy to use
+- It allows you to run containers in different regions
 - It is typically the cheapest; since no virtual machines or higher-level services are required, **you only pay for the memory and CPU used by your container group while it is active.**
 - Unlike the [DockerOperator](https://airflow.apache.org/docs/apache-airflow/1.10.4/_api/airflow/operators/docker_operator/index.html), it does not require running a container on the host machine
 
--- a/guides/airflow-azure-container-instances.md
+++ b/guides/airflow-azure-container-instances.md
@@ -10,7 +10,7 @@ tags: ["Integrations", "Azure", "DAGs"]
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+Note that this operator can also be used to run existing container instances and make certain updates, including the docker image, environment variables, or commands. Some updates to existing container groups are not possible with the operator, including CPU, memory, and GPU; these updates require deleting the existing container group and recreating it, which can be accomplished using the [AzureContainerInstanceHook](https://github.com/apache/airflow/blob/master/airflow/providers/microsoft/azure/hooks/azure_container_instance.py).
+
+### When to use ACI
+
+There are multiple ways to work with containers with Airflow on Azure, each of which could be the right fit from your use case. The most flexible and scalable method would be to the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow/stable/kubernetes.html). This will not only let you run any container with Airflow, but also run it as a Kubernetes pod, which allows you to pass in resource requests and other Kubernetes native parameters. However, it requires an [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/) cluster (or a hand rolled Kubernetes cluster).
```
jwitz, line 31:
```suggestion
There are multiple ways to manage containers with Airflow on Azure. The most flexible and scalable method is to use the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow/stable/kubernetes.html). This lets you run any container as a Kubernetes pod, which means you can pass in resource requests and other native Kubernetes parameters. Using this operator requires an [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/) cluster (or a hand-rolled Kubernetes cluster).
```

jwitz: `you can pass in resource requests and other native Kubernetes parameters.` To where? The DAG?

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+If you are not running on [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/), ACI can be a great choice:
+
+- It requires very little setup, and is easy to use
+- It allows you to run containers in different regions
```
jwitz, line 36:
```suggestion
- You can run containers in different regions
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+### When to use ACI
+
+There are multiple ways to work with containers with Airflow on Azure, each of which could be the right fit from your use case. The most flexible and scalable method would be to the [KubernetesPodOperator](https://airflow.apache.org/docs/apache-airflow/stable/kubernetes.html). This will not only let you run any container with Airflow, but also run it as a Kubernetes pod, which allows you to pass in resource requests and other Kubernetes native parameters. However, it requires an [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/) cluster (or a hand rolled Kubernetes cluster).
+
+If you are not running on [AKS](https://azure.microsoft.com/en-us/services/kubernetes-service/), ACI can be a great choice:
+
+- It requires very little setup, and is easy to use
```
jwitz, line 35:
```suggestion
- It's easy to use and requires little setup
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- It allows you to run containers in different regions
+- It is typically the cheapest; since no virtual machines or higher-level services are required, **you only pay for the memory and CPU used by your container group while it is active.**
```
jwitz, line 37:
```suggestion
- It's typically the cheapest; since no virtual machines or higher-level services are required, **you only pay for the memory and CPU used by your container group while it is active**
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- Unlike the [DockerOperator](https://airflow.apache.org/docs/apache-airflow/1.10.4/_api/airflow/operators/docker_operator/index.html), it does not require running a container on the host machine
+
+With this in mind, we would recommend using ACI with the AzureContainerInstancesOperator for testing or for light weight tasks that don't require scaling. For heavy production workloads, we would recommend sticking with AKS and the KubernetesPodOperator.
```
jwitz, line 40:
```suggestion
With these points in mind, we recommend using ACI with the AzureContainerInstancesOperator for testing or lightweight tasks that don't require scaling. For heavy production workloads, we recommend sticking with AKS and the KubernetesPodOperator.
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Example
+
+Using Airflow to create and run an ACI is straight forward. First, identify the Azure resource group you want to create the ACI in (or create a new one); then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, see this [link](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
```
jwitz, line 44:
```suggestion
Using Airflow to create and run an ACI is straightforward: You first identify the Azure resource group you want to create the ACI in (or create a new one), then ensure your Azure instance has a service principle with write access over that resource group. For more information on setting this up, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-create-service-principal-portal).
```

jwitz: We introduce "ACI" as meaning a singular "Azure Container Instances" service, but then refer to it here as "an ACI". Should we be saying "run an Azure container instance" instead of "Run an ACI"? This is how Microsoft refers to it themselves: https://docs.microsoft.com/en-us/azure/container-instances/container-instances-quickstart

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+Next, create an Airflow connection with the type `Azure Container Instance`, and fill it out with your Client ID as the login, Client Secret as the password, and Tenant ID and Subscription ID in the extra's field as json. It should look something like this:
```
jwitz, line 46:
```suggestion
Next, create an Airflow connection with the type `Azure Container Instance`. Specify your Client ID in the Login field, Client Secret in the Password field, and Tenant and Subscription IDs in the Extras field as json. It should look something like this:
```

jwitz: Assuming these are all fields, feel free to ignore this if they're not!

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+![ACI Connection](https://assets2.astronomer.io/main/guides/azure-container-instances/aci_connection.png)
+
+Next, define a DAG using the AzureContainerInstancesOperator:
```
jwitz, line 50:
```suggestion
Lastly, define a DAG using the AzureContainerInstancesOperator:
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+        cpu=1,
+        memory_in_gb=1.5,
+        fail_if_exists=False
+
+    )
+```
+
+The parameters for the operator are:
+
+- **ci_conn_id:** the connection ID for the connection we created above
```
jwitz, line 92:
```suggestion
- **ci_conn_id:** The connection ID for the Airflow connection we created above
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **registry_conn_id:** the connection ID to connect to a registry. In this case we used DockerHub, which is public and does not require credentials, so we pass in `None`
```
jwitz, line 93:
```suggestion
- **registry_conn_id:** The connection ID to connect to a registry. In this case we use DockerHub, which is public and does not require credentials, so we pass in `None`
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **resource_group:** our Azure resource group
```
jwitz, line 94:
```suggestion
- **resource_group:** Our Azure resource group
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **name:** the name we want to give our ACI. Note that this must be unique within the resource group
```
jwitz, line 95:
```suggestion
- **name:** The name we want to give our ACI. Note that this must be unique within the resource group
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **image:** the docker image we want to use for the container. In this case we use a simple Hello World example from Docker
```
jwitz, line 96:
```suggestion
- **image:** The Docker image we want to use for the container. In this case we use a simple Hello World example from Docker
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **region:** the Azure region we want our ACI deployed to
```
jwitz, line 97:
```suggestion
- **region:** The Azure region we want our ACI deployed to
```
"Azure container instace" instead of "ACI"?

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **CPU:** the number of CPUs to allocate to the container. In this case we have used the default minimum for example. For more information on allocating CPU and memory, see this [link](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-faq).
```
jwitz, line 98:
```suggestion
- **CPU:** The number of CPUs to allocate to the container. In this example we use the default minimum. For more information on allocating CPUs and memory, refer to the [Azure documentation](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-faq).
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **memory_in_gb**: the amount of memory to allocate to the container. In this case we have used the default minimum for example.
```
jwitz, line 99:
```suggestion
- **memory_in_gb**: The amount of memory to allocate to the container. In example we use the default minimum.
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+- **fail_if_exists:** whether we want the the operator to raise an exception if the container group already exists (defaults to True). If set to False and the container group name already exists within the given resource group, the operator will attempt to update the container group based on the other parameters before running and terminating upon completion.
```
jwitz, line 100:
```suggestion
- **fail_if_exists:** Whether we want the the operator to raise an exception if the container group already exists (default value is True). If it's set to False and the container group name already exists within the given resource group, the operator will attempt to update the container group based on the other parameters before running and terminating upon completion
```

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+Note that you can also provide the operator with other parameters such as environment variables, volumes, and a command as needed to run the container.
```
jwitz, line 102:
```suggestion
Note that you can also provide the operator with other parameters, such as environment variables or volumes, to run the container as needed.
```
Sorry the "a command" was bugging me - feel free to keep it in if it truly is only one command you can add!

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+Now if we run this DAG, an ACI will spin up, run the container with the Hello World image, and then spin down. If we look at the Airflow task log, we see the printout from the container has propagated to the logs:
```
jwitz, line 104:
```suggestion
If we run this DAG, an ACI will spin up, run the container with the Hello World image, and spin down. If we look at the Airflow task log, we see the printout from the container has propagated to the logs:
```

jwitz (changes requested):
Overall looks really good @kentdanas ! Only uncommitable suggestions I made were to:

* Link out to more Microsoft docs when possible
* Standardize use of "ACI" to refer to the service, "Azure container instance" to refer to the actual instances
* Moving the note from the beginning of the document

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
+---
+title: "Orchestrating Azure Container Instances with Airflow"
+description: "Orchestrating containers with Azure Container Instances from your Apache Airflow DAGs."
+date: 2020-12-21T00:00:00.000Z
+slug: "airflow-azure-container-instances"
+tags: ["Integrations", "Azure", "DAGs"]
+---
+
+> Note: All code in this guide can be found in [this Github repo](https://github.com/astronomer/azure-operator-tutorials).
```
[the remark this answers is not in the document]
kentdanas: I'll defer to @petedejoy on that one since we have this as a pattern in some of our other guides. In some ways I think it's helpful to point out the code right up front, but I do understand your point.

guides/airflow-azure-container-instances.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Orchestrating ACI with Airflow
+
+### The Azure Container Instances Operator
+
+The easiest way to orchestrate Azure Container Instances with Airflow is to use the [AzureContainerInstancesOperator](https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/contrib/operators/azure_container_instances_operator.html). This operator starts a container on ACI, runs the container, and terminates the container when all processes are completed. 
+
+The only prerequisites for using this operator are:
+
+- An Azure account with a resource group created
```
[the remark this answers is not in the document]
kentdanas: I would expect them to be familiar with these concepts since the assumption is that this would be used by people who are already using Azure. But, definitely can't hurt to have more msft docs linked; I'll add some.

guides/airflow-azure-container-instances.md
kentdanas: To the operator

guides/airflow-azure-container-instances.md
kentdanas: Yep good call, I'll update

guides/airflow-azure-container-instances.md
kentdanas: The param you pass in is actually called "command". Agree it sounds weird, I just couldn't think of a different way to phrase it. Running a container with a command is a pretty common use case so thought it was worth calling out.

jwitz (approved):
Looks good, thanks for writing this @kentdanas ! Excited to read it once it's published.

PR #184, opened 2021-01-28
Replaced references to max_threads in two guides

jwitz:
Closes https://github.com/astronomer/docs/issues/213 .

In Airflow 1.10.14, the max_threads variable was renamed to parsing_processes. This PR replaces references to max_threads in Airflow guides where it's referenced. 

I was unsure of how much support we want to offer for previous Airflow versions in guides, so as a precaution I included notes explaining the change (For anyone still on 1.10.13). Feel free to axe these if you don't think they're necessary.

While I was in here, I also did a cheeky title update: When a preposition is part of a phrasal verb (e.g. working out, breaking down), it functions as part of the verb and should be capitalized.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -1616,3 +1616,4 @@ mongo_to_redshift.py
  - guides/airflow-databricks.md
 parameterizing
 granularly
+parsing_processes
--- a/.spelling
+++ b/.spelling
@@ -1606,9 +1606,14 @@ mongo_to_redshift.py
 parameterizing
 granularly
 AzureContainerInstancesOperator
+AzureDataExplorerQueryOperator
 ci\_conn\_id
 DockerHub
 memory\_in\_gb
 parsing_processes
 max_threads
 ConnType
+auth_method
+AzureContainerInstanceHook
+Covid-19
+david-astro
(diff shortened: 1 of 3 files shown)


petedejoy:
@jwitz looks like the spellcheck CI caught some issue here.

jwitz:
> @jwitz looks like the spellcheck CI caught some issue here.

@petedejoy If you look in the CI report, the spell check is getting caught on words like AzureContainerInstancesOperator and setting names. We have this problem in `docs` as well, and resolved to collect these words over time and add them in a batch PR to the CI files. In the meantime though, I don't think there's anything I can do to resolve this check without PRing to your pipeline.

jwitz:
@petedejoy I tried adding the words that were being caught here to `.spelling`, but it seems like the pipeline is still getting caught on them. What's the proper way to get this check to pass?

petedejoy:
Hmm yeah this is weird- looks like it's getting hung up on that `AzureContainerInstancesOperator`, even though it's in `.spelling`. @sunkickr - any idea what's up with this?

petedejoy (approved):

PR #207, opened 2021-05-13
Update DAG writing best practices guide

kentdanas:
Hey @jwitz! Did a pretty major overhaul of this guide (one of our more popular ones). Would love if you could work your magic on it. Thank you as always!!!

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -954,6 +954,7 @@ stderr
 backfilling
 timebased
 kedro
+idempotence
  - posts/4-pillars-to-becoming-data-driven.md
 astronomer.io
 _
--- a/.spelling
+++ b/.spelling
@@ -1491,6 +1491,7 @@ depends_on_past
 wait_for_downstream
 - guides/dynamically-generating-dags.md
 Create_DAG
+ELT-type
 - guides/error-notifications-in-airflow.md
 __`e
 ____
@@ -1636,3 +1637,5 @@ AwsBaseHook
 BaseSensorOperator
 PostgresHook
 SSHHook
+ - guides/airflow-scaling-workers.md
+default_airflow.cfg
--- a/.spelling
+++ b/.spelling
@@ -957,7 +957,6 @@ kedro
 idempotence
  - posts/4-pillars-to-becoming-data-driven.md
 astronomer.io
-default_airflow.cfg
  - posts/why-we-built-our-data-platform-on-aws-and-why-we-rebuilt-it-with-open-source.md
 checkpointing
  - posts/why-is-my-data-playing-hard-to-get.md
--- a/.spelling
+++ b/.spelling
@@ -1627,6 +1627,9 @@ parsing_processes
 deserialization
 returned_value
 dbt
+serializable
+versioning
+deserializing
 AwsLambdaHook
 AwsBaseHook
 BaseSensorOperator
--- a/.spelling
+++ b/.spelling
@@ -70,7 +70,7 @@ autoscaler
 airflow
 airflowers
 iberman
-boolean 
+boolean
 astronomerio
 dbaas
 datahub
@@ -917,7 +917,7 @@ Mika
 Heino
 Solita
 VideoAmp
-How-to's 
+How-to's
 Peyman
 Mohajerian
 Vinayak
@@ -1494,14 +1494,14 @@ Create_DAG
 __`e
 ____
 - guides/facebook-ads-to-redshift.md
-age_gender 
+age_gender
 device_platform
 region_country
 no_breakdown
 account_id
 ad_id
 adset_id
-ad_name 
+ad_name
 adset_name
 campaign_id
 date_start
@@ -1520,7 +1520,7 @@ facebook_ads_to_redshift.py
 v18.09
 - guides/github-to-redshift.md
 issue_comments
-pull_requests 
+pull_requests
 github_to_redshift.py
 - guides/google-analytics-to-redshift.md
 v4
@@ -1627,3 +1627,8 @@ parsing_processes
 deserialization
 returned_value
 dbt
+AwsLambdaHook
+AwsBaseHook
+BaseSensorOperator
+PostgresHook
+SSHHook
--- a/.spelling
+++ b/.spelling
@@ -957,7 +957,7 @@ kedro
 idempotence
  - posts/4-pillars-to-becoming-data-driven.md
 astronomer.io
-_
+default_airflow.cfg
  - posts/why-we-built-our-data-platform-on-aws-and-why-we-rebuilt-it-with-open-source.md
 checkpointing
  - posts/why-is-my-data-playing-hard-to-get.md
--- a/guides/dag-best-practices.md
+++ b/guides/dag-best-practices.md
@@ -46,7 +46,7 @@ today = datetime.today()
 yesterday = datetime.today() - timedelta(1)
 ```
 
-The example above creates date variables that can be used by tasks by calling `datetime` package functions. If this code is in the DAG file, these functions will be executed on every scheduler heartbeat, which may not be performant. Even more importantly, this doesn't produce an idempotent DAG; if you needed to rerun a previously failed DAG Run for a past date, you wouldn't be able to do so with this implementation because the date variable is relative to the actual date, not the DAG execution date. 
+If this code is in a DAG file, these functions will be executed on every Scheduler heartbeat, which may not be performant. Even more importantly, this doesn't produce an idempotent DAG: If you needed to rerun a previously failed DAG Run for a past date, you wouldn't be able to because `datetime.today()` is relative to the current date, not the DAG execution date. 
 
 A better way of implementing this is by using an Airflow variable:
 
--- a/guides/dag-best-practices.md
+++ b/guides/dag-best-practices.md
@@ -16,7 +16,7 @@ In general, most of the best practices we cover here fall into two categories: D
 
 Before we jump into best practices specific to Airflow, we'll call out one, idempotency, which applies to all data pipelines. Idempotency is very important, and you may notice that other best practices in this guide are in part to support this one.
 
-[Idempotency]((https://en.wikipedia.org/wiki/Idempotence)) is the concept whereby an operation can be repeated multiple times without changing the result. A common example is a crosswalk button; no matter how many times you push the button, you're going to get the same result. With a DAG, this translates to the same DAG, if run multiple times, will generate the same results. Applying this concept will make recovery from any failures in your DAGs easier, and ensure data loss is prevented.
+[Idempotency](https://en.wikipedia.org/wiki/Idempotence) is the concept whereby an operation can be repeated multiple times without changing the result. A common example is a crosswalk button; no matter how many times you push the button, you're going to get the same result. With a DAG, this translates to the same DAG, if run multiple times, will generate the same results. Applying this concept will make recovery from any failures in your DAGs easier, and ensure data loss is prevented.
 
 ## DAG Design
 
@@ -251,6 +251,6 @@ Changing the name of a DAG also creates a new entry in the database, which power
 
 ### Set Retries at the DAG Level
 
-Even if your code is perfect, failures happen. In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens you may see Airflow's logs mention a [zombie process]((https://en.wikipedia.org/wiki/Zombie_process)).
+Even if your code is perfect, failures happen. In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens you may see Airflow's logs mention a [zombie process](https://en.wikipedia.org/wiki/Zombie_process).
 
 Issues like this can be resolved by using task retries. Best practice is to set retries as a `default_arg` so they are applied at the DAG level, and get more granular for specific tasks only where necessary. A good range to try is ~2–4 retries.
--- a/guides/dag-best-practices.md
+++ b/guides/dag-best-practices.md
@@ -31,7 +31,9 @@ The following DAG design principles will help to make your DAGs idempotent, effi
 
 ### Keep Tasks Atomic
 
-When breaking up your pipeline into individual tasks, ideally each task should be atomic. This means each task should be responsible for one operation that can be re-run independently of the others, which supports idempotency.
+When breaking up your pipeline into individual tasks, ideally each task should be atomic. This means each task should be responsible for one operation that can be re-run independently of the others. Said another way, in an atomized a task, a success in part of the task means a success of the entire task. 
+
+For example, in an ETL pipeline you would ideally want your Extract, Transform, and Load operations covered by three separate tasks. Atomizing these tasks allows you to rerun each operation in the pipeline independently, which supports idempotence.
 
 ### Use Template Fields, Variables, and Macros
 
--- a/guides/dag-best-practices.md
+++ b/guides/dag-best-practices.md
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
 tags: ["DAGs", "Best Practices", "Basics", "Templating", "Tasks"]
 ---
 
-Welcome to our guide on writing Airflow DAGs. In this piece, we'll walk through some high-level concepts involved in Airflow DAGs, explain what to stay away from, and cover some useful tricks that will hopefully be helpful to you.
+Writing Airflow DAGs is typically a pretty easy skill to pick up if you know a little bit of Python; you just need to instantiate your DAG and your operators and you're off and running. However, writing optimized DAGs that make the most of Airflow can sometimes feel like more of an art than a science. In this guide, we'll cover some best practices when writing DAGs that will help ensure your pipelines are efficient and sustainable.
 
-If you're interested in further DAG writing help or general Airflow assistance, we offer support packages that give you on-demand access to Airflow experts. [Drop us a line](https://www.astronomer.io/#request) if you'd like to chat.
+In general, most of the best practices we cover here fall into two categories: DAG design, and using Airflow as an orchestrator. For an in-depth walk through and examples of some of the concepts covered here, check out our DAG writing best practices [webinar recording](https://www.astronomer.io/blog/dag-writing-best-practices-in-apache-airflow), and our [Github repo](https://github.com/astronomer/webinar-dag-writing-best-practices) with good and bad example DAGs.
 
-### Idempotency
 
-Data pipelines are a messy business with a lot of various components that can fail. [Idempotent](https://en.wikipedia.org/wiki/Idempotence) DAGs allow you to deliver results faster when something breaks and can save you from losing data down the road.
+## Idempotency
 
-### Use Retries
+Before we jump into best practices specific to Airflow, we'll call out one, idempotency, which applies to all data pipelines. Idempotency is very important, and you may notice that other best practices in this guide are in part to support this one.
 
-In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens you may see Airflow's logs mention a zombie process.
+[Idempotency]((https://en.wikipedia.org/wiki/Idempotence)) is the concept whereby an operation can be repeated multiple times without changing the result. A common example is a crosswalk button; no matter how many times you push the button, you're going to get the same result. With a DAG, this translates to the same DAG, if run multiple times, will generate the same results. Applying this concept will make recovery from any failures in your DAGs easier, and ensure data loss is prevented.
 
-A [zombie process](https://en.wikipedia.org/wiki/Zombie_process) occurs when Airflow goes to check on the process for a task that it thinks is running but finds out that the process was killed or is otherwise not actually running. (It could have been killed for any number of reasons.)
+## DAG Design
 
-This can often be resolved by bumping up retries on the task. A good range to try is ~2–4 retries.
+The first category of best practices specific to Airflow relate to DAG design. These are practices that will help ensure your DAGs are idempotent, efficient, and readable.
+
+### Keep Tasks Atomic
+
+When breaking up your pipeline into individual tasks, ideally each task should be atomic. This means each task should be responsible for one operation that can be re-run independently of the others, which supports idempotency.
+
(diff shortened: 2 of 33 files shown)


guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+            JOIN today_covid_data b
+            ON a.state=b.state
+            JOIN two_day_rolling_avg c
+            ON a.state=b.two_day_avg;''',
+            params={'today': today, 'yesterday':yesterday}
+    )
+```
 
-Depending on your data retention policy you could modify the load logic and re-run the entire historical pipeline without having to re-run the extracts. This is also useful in situations where you no longer have access to the source system (i.e. hit an API limit).
+Instead, you would want to keep that SQL code in a separate file (`covid_state_query.sql` below), and call it into your DAG like this:
```
jwitz, line 140:
```suggestion
To avoid that SQL code running on every Scheduler heartbeat, you would want to keep it in a separate file and call it into your DAG. In the following example, we call in a file named `covid_state_query.sql` into our PostgresOperator instantiation, which embodies the best practice:
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+Airflow executes all code in the `DAGS_Folder` on every scheduler heartbeat, so from a technical perspective even a small amount of top level code can cause performance issues. You especially don't want top level code that is making requests to external systems, like an API or a database, or any function calls outside of your tasks. Additionally, lots of top level code in your DAG files makes them harder to read, maintain, update, and debug.
 
-If you are using s3 as your intermediary, it is best to set a policy restricted to a dedicated s3 bucket to use in your Airflow s3 connection object. This policy will need to read, write, and delete objects.
+Try to treat your DAG file like a config file and leave all the heavy lifting for hooks and operators. If your DAGs need to access additional code such as a SQL script or a Python function, keep that code in a separate file that can be read into your DAG.
```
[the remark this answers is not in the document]
jwitz: In general, I was confused in this section when trying to understand the difference between the DAG, the DAG instantiation, the DAG file, and the DAG run. I think the instantiation is the most useful way to understand this best practice, so maybe focus on it by saying something like "All of the heavy lifting should occur only within your DAG and Operator instantiations."

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+            params={'state': "'" + state + "'"}
+        )
+```
 
-For example, the `s3_key` and `since` and `until` fields are set as `template_fields` here:
- `https://github.com/airflow-plugins/google_analytics_plugin/blob/master/operators/google_analytics_reporting_to_s3_operator.py#L41`
+### Use a Consistent Method for Task Dependencies
 
- This allows for these values to be dynamically set by the `schedule_interval`.
+In Airflow, task dependencies can be set multiple ways. You can use `set_upstream()` and `set_downstream()` functions, or use `<<` or `>>` operators. Which method you use is a matter of personal preference, but for readability it's best practice to choose a consistent method.
```
jwitz, line 177:
```suggestion
In Airflow, task dependencies can be set multiple ways. You can use `set_upstream()` and `set_downstream()` functions, or you can use `<<` and `>>` operators. Which method you use is a matter of personal preference, but for readability it's best practice to choose one method and stick with it.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+task_1 >> task_2 >> [task_3, task_4]
+```
 
-A dynamic start_date is misleading. It can cause failures when clearing out failed task instances and missing DAG runs.
 
-## Transformations
+## Use Airflow as an Orchestrator
 
-Look to implement an ELT (extract, load, transform) data pipeline pattern with your DAG definition file. This means that you should look to offload as much of the transformation logic to the source systems or the destinations systems as possible. With python at your fingertips it can be tempting to attempt the transformations in the DAG but offloading those transformations to the source or destination systems will lead to better overall performance and keeps your DAG lean and readable.
+The next category of best practices is to use Airflow as an orchestrator; this will allow you to make the most of Airflow as it was designed, and ensure it can scale to meet your needs. Sticking to the following best practices will help ensure you don't experience common pitfalls using Airflow as the right tool for the wrong job.
```
jwitz, line 196:
```suggestion
The next category of best practices relates to using Airflow as what it was originally designed to be: an orchestrator. Using Airflow as an orchestrator makes it easier to scale and pull in the right tools based on your needs.
```
I wasn't sure what the second sentence originally meant, so I tried to frame it using positive verbs. As it was before, there was too many double-negatives.

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
 
-### Use Staging Tables
+### Make Use of Provider Packages
 
-Try to use staging tables before pushing to a final destination. This makes debugging errors easier as you'll have the exact data that caused an error and adds a layer of safety.
+A good way to ensure you're using Airflow as it was designed is to make use of [provider packages](https://airflow.apache.org/docs/apache-airflow-providers/) to orchestrate jobs with other tools. One of the best things about Airflow is its robust and active community, which has resulted in many integrations between Airflow and other tools in the data ecosystem. Wherever possible, it's best practice to make use of these existing integrations. This allows for easier Airflow adoption for teams that may already be using existing tools, and means you have to write less code since many existing hooks and operators have taken care of that for you.
```
jwitz, line 200:
```suggestion
One way you can use Airflow as an orchestrator is by making use of existing [provider packages](https://airflow.apache.org/docs/apache-airflow-providers/), which orchestrate jobs using third party tools. One of the best aspects of Airflow is its robust and active community, which has resulted in many integrations between Airflow and other tools in the data ecosystem. Wherever possible, it's best practice to make use of these integrations. This makes it easier for teams using existing tools to adopt Airflow, and it means you get to write less code since many existing hooks and operators have taken care of that for you.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
 
-**Note** By default, each task counts as its own database session, so avoid temporary tables that only last a session. Instead, have the last task in your DAG clear out intermediary tables if everything runs successfully.
+For easy discovery of all the great provider packages out there, check out the [Astronomer Registry](https://registry.astronomer.io/).
 
-### Mongo Source
+### Don't Use Airflow as a Processing Framework
 
-Use [aggregation pipelines](https://docs.mongodb.com/manual/core/aggregation-pipeline/) to perform your transformations on extract from a Mongo source.
+Airflow was not designed to be a processing framework. Since DAGs are written in Python, you have the power of Python behind you and it can be tempting to make use of data processing libraries like Pandas. However, processing large amounts of data within your Airflow tasks will not scale well, and should only be used in cases where the data are limited in size. A better option for scalability is to offload any heavy duty processing to a framework like [Apache Spark](https://spark.apache.org/), and use Airflow to orchestrate those jobs.
```
jwitz, line 206:
```suggestion
Conversely, Airflow was _not_ designed to be a processing framework. Since DAGs are written in Python, it can be tempting to make use of data processing libraries like Pandas. However, processing large amounts of data within your Airflow tasks does not scale and should only be used in cases where the data are limited in size. A better option for scalability is to offload any heavy duty processing to a framework like [Apache Spark](https://spark.apache.org/), then use Airflow to orchestrate those jobs.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
-Try to do basic transformations and aggregations in SQL queries - this offloads transformation logic onto the source system and keeps your DAG readable.
+- Ensure your Airflow infrastructure has the necessary resources.
+- Use the Kubernetes Executor to isolate task processing and have more control over resources at the task level.
+- Use a [custom XCom backend](https://www.astronomer.io/guides/custom-xcom-backends) if you need to pass any data between the tasks so you don't overload your metadata database.
 
-## Readability
+### Use Intermediary Data Storage
 
-### Use a consistent file structure
+It can be tempting to write your DAGs so that they move data directly from your source to destination. It usually makes for less code and involves fewer pieces, but doing so removes your ability to re-run just the extract or load portion of the pipeline individually. By putting an intermediary storage layer such as S3 or SQL Staging tables in between your source and destination, you can separate the testing and re-running of the extract and load.
```
jwitz, line 216:
```suggestion
Because it requires less code and fewer pieces, it can be tempting to write your DAGs to move data directly from your source to destination. However, this means you can't individually rerun the extract or load portions of the pipeline. By putting an intermediary storage layer such as S3 or SQL Staging tables in between your source and destination, you can separate the testing and rerunning of the extract and load.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
 
-To keep any custom plugins easy for someone else to use, use a consistent file structure. At Astronomer, we use:
+Depending on your data retention policy, you could modify the load logic and re-run the entire historical pipeline without having to re-run the extracts. This is also useful in situations where you no longer have access to the source system (e.g. hit an API limit).
```
jwitz, line 218:
```suggestion
Depending on your data retention policy, you could modify the load logic and rerun the entire historical pipeline without having to rerun the extracts. This is also useful in situations where you no longer have access to the source system (e.g. you hit an API limit).
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
-│   ├── __init__.py
-│   └── hook_one.py
-└── operators  <--- Contains the operators
-    ├── __init__.py
-    └── operator_one.py
-```
+### Use an ELT Framework
 
-See [here](https://github.com/airflow-plugins/) for examples!
+Whenever possible, look to implement an ELT (extract, load, transform) data pipeline pattern with your DAGs. This means that you should look to offload as much of the transformation logic to the source systems or the destinations systems as possible, and is one way of ensuring you aren't using Airflow as your processing framework. Many modern data warehouse tools, such as [Snowflake](https://www.snowflake.com/), give you easy to access to compute to support the ELT framework, and are easily used in conjunction with Airflow.
```
jwitz, line 222:
```suggestion
Whenever possible, look to implement an ELT (extract, load, transform) data pipeline pattern with your DAGs. This means that you should look to offload as much of the transformation logic to the source systems or the destinations systems as possible, which avoids using Airflow as a processing framework. Many modern data warehouse tools, such as [Snowflake](https://www.snowflake.com/), give you easy to access to compute to support the ELT framework, and are easily used in conjunction with Airflow.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+## Other Best Practices
 
-Changing the name of a DAG also creates a new entry in the database, which powers the dashboard, so follow a consistent naming convention since changing a DAG's name doesn't delete the entry in the database for the old name.
+Finally, here are a few other noteworthy best practices that don't fall under the two categories above.
 
-### Avoid top level code in your DAG file
+### Use a Consistent File Structure
 
-The Airflow executor executes top level code on every heartbeat, so a small amount of top level code can cause performance issues. Try to treat the DAG file like a config file and leave all the heavy lifting for the hook and operator.
+Having a consistent file structure for Airflow projects can help keep things organized and easy to adopt when working with Airflow both within teams and across teams. At Astronomer, we use:
```
jwitz, line 231:
```suggestion
Having a consistent file structure for Airflow projects keeps things organized and easy to adopt. At Astronomer, we use:
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+Changing the name of a DAG also creates a new entry in the database, which powers the dashboard, so follow a consistent naming convention since changing a DAG's name doesn't delete the entry in the database for the old name.
 
-```python
-task_1.set_downstream(task_2)
-task_2.set_downstream(task_3)
-```
+### Set Retries at the DAG Level
 
-or this
+Even if your code is perfect, failures happen. In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens you may see Airflow's logs mention a [zombie process](https://en.wikipedia.org/wiki/Zombie_process).
```
jwitz, line 254:
```suggestion
Even if your code is perfect, failures happen. In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens, you might see Airflow's logs mention a [zombie process](https://en.wikipedia.org/wiki/Zombie_process).
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
 
-```python
-task_3 >> task_2
-task_2 >> task_1
-```
+Issues like this can be resolved by using task retries. Best practice is to set retries as a `default_arg` so they are applied at the DAG level, and get more granular for specific tasks only where necessary. A good range to try is ~2–4 retries.
```
jwitz, line 256:
```suggestion
Issues like this can be resolved by using task retries. Best practice is to set retries as a `default_arg` so they are applied at the DAG level and get more granular for specific tasks only where necessary. A good range to try is ~2–4 retries.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+└── requirements.txt # For any Python packages
 
-Task dependencies are set using `set_upstream()` and `set_downstream()`. Using either will depend on your preferences, but it is best to stay consistent with which one you use.
+```
 
-#### Example
+### Use DAG Name and Start Date Properly
 
-Instead of this
+You should always use a static `start_date` with your DAGs. A dynamic start_date is misleading, and can cause failures when clearing out failed task instances and missing DAG runs.
```
jwitz, line 246: This kind of sound like an idempotence issue?

jwitz (changes requested):
Sorry this took a minute @kentdanas but hopefully some of the suggestions I make here are useful! I'll admit that some of my questions stem from a lack of actual Python/ DAG authorship, but my guess is that others reading this guide will be coming from a similar place.

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+## Idempotency
 
-### Use Retries
+Before we jump into best practices specific to Airflow, we'll call out one, idempotency, which applies to all data pipelines. Idempotency is very important, and you may notice that other best practices in this guide are in part to support this one.
 
-In a distributed environment where task containers are executed on shared hosts, it's possible for tasks to be killed off unexpectedly. When this happens you may see Airflow's logs mention a zombie process.
+[Idempotency](https://en.wikipedia.org/wiki/Idempotence) is the concept whereby an operation can be repeated multiple times without changing the result. A common example is a crosswalk button; no matter how many times you push the button, you're going to get the same result. With a DAG, this translates to the same DAG, if run multiple times, will generate the same results. Applying this concept will make recovery from any failures in your DAGs easier, and ensure data loss is prevented.
```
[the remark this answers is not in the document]
kentdanas: Yep, you're right, it's at the DAG Run level. The DAG itself won't always produce the same results, because the data, time, etc. can change, but each **DAG Run** should always be idempotent. e.g. DAG Run A should always process data set A for date x/x/xxxx, and DAG Run B should always process data for data set b and date y/y/yyyy. Changed the wording a little bit, does that make more sense?

kentdanas, line 19:
```suggestion
In the context of Airflow, a DAG is considered idempotent if every DAG Run generates the same results even when run multiple times. Designing idempotent DAGs decreases recovery time from failures and prevents data loss.
```

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
 
-A [zombie process](https://en.wikipedia.org/wiki/Zombie_process) occurs when Airflow goes to check on the process for a task that it thinks is running but finds out that the process was killed or is otherwise not actually running. (It could have been killed for any number of reasons.)
+## DAG Design
+
+The first category of best practices specific to Airflow relate to DAG design. These are practices that will help ensure your DAGs are idempotent, efficient, and readable.
+
+### Keep Tasks Atomic
+
+When breaking up your pipeline into individual tasks, ideally each task should be atomic. This means each task should be responsible for one operation that can be re-run independently of the others, which supports idempotency.
```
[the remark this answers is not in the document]
kentdanas: For your own info, you can re-run just part of a DAG in Airflow....BUT, that's not what we're talking about here (sorry if this is more confusing :) ). Atomicity is referring to the content of each task, not the content of the DAG or how all the tasks fit together. Another way of saying it is that a task is atomic when if one part of the task fails, the entire task fails, or if one part succeeds, the entire task succeeds. As a more practical example, say I have a DAG that grabs data from an API endpoint and then loads it into a database. If I designed this DAG as one task that did everything, it would not be atomic; if the API request was successful but the database load failed, one "part" of the task succeeded, but the task as a whole did not succeed. If I split the DAG into two tasks, then it would be atomic. In vague terms, I usually think of it as each task does a "single thing" as opposed to multiple things. Sometimes atomicity isn't possible, but it's good to strive for it. 

Does that help? I'll think about ways to re-word this better to make it more clear, I definitely struggle with articulating this one in a couple of sentences without just talking through an example.

guides/dag-best-practices.md
```diff
@@ -7,169 +7,250 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/bestwritingpract
...
+For example, if your DAG needs to be date parameterized you should use an Airflow date variable rather than a Python function that determines the date.
+
+```python
+# Variables used by tasks
+# Bad example - Define today's and yesterday's date using datetime module
+today = datetime.today()
+yesterday = datetime.today() - timedelta(1)
+```
+
+The example above creates date variables that can be used by tasks by calling `datetime` package functions. If this code is in the DAG file, these functions will be executed on every scheduler heartbeat, which may not be performant. Even more importantly, this doesn't produce an idempotent DAG; if you needed to rerun a previously failed DAG Run for a past date, you wouldn't be able to do so with this implementation because the date variable is relative to the actual date, not the DAG execution date.
```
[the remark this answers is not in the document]
kentdanas: Yep, it should be the same result for a particular DAG Run, not necessarily every time you "run" the DAG. (confusing talking about running the DAG and a DAG Run as two different things, but that is the Airflow terminology....). Tried to clarify that above.

guides/dag-best-practices.md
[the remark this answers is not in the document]
kentdanas: The main point of the "heavy lifting" comment is that you should be using hooks and operators (even if you define them yourself) as the "meat" of the code, as opposed to writing a bunch of python in the DAG file. The hook and operator code would live in separate files, that are then referenced in the DAG file. So the DAG and operator instantiations (which are just calling Airflow's `DAG` function, and whatever operator functions) need to happen in the DAG file, but that's ideally _all_ you want happening in the DAG file.

guides/dag-best-practices.md
kentdanas: Your re-write is perfect, thank you for fixing my nonsense :)

guides/dag-best-practices.md
kentdanas: It could be!n If any tasks in your DAG were based on the start date, or if you had DAG runs that were older than a new changed start date.

kentdanas:
All good, thanks @jwitz! Tried to answer your questions and clarify a few things. Let me know if that helps!

guides/dag-best-practices.md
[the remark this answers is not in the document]
jwitz: New wording clears this up!

guides/dag-best-practices.md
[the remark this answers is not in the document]
jwitz: I think an ETL pipeline is a useful example here. Maybe just a single sentence like "For example, in an ETL pipeline you would want your Extract, Transform, and Load operations covered by three separate tasks. Atomizing these tasks allows you to rerun each operation in the ETL independently, which supports idempotence."

guides/dag-best-practices.md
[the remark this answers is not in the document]
jwitz: I see your clarification above, but I'm not seeing seeing the changes in this suggestion, were those committed?

jwitz (approved):
Left one minor suggestion with regards to atomized tasks and another comment about an existing suggestions, but otherwise looks great to me @kentdanas ! Quite the improvement on the original guide 🔥

guides/dag-best-practices.md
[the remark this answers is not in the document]
kentdanas: Oh sorry, might have confused things here. When I said clarified above, I meant in the section above where we initially talked about idempotence. For this section was just going to commit your suggested changes, unless there was something else you think I should add?

.spelling
```diff
@@ -954,6 +954,7 @@ stderr
 backfilling
 timebased
 kedro
+idempotence
  - posts/4-pillars-to-becoming-data-driven.md
 astronomer.io
 _
```
kentdanas, line 960:
```suggestion
default_airflow.cfg file
```

.spelling
```diff
@@ -954,9 +954,10 @@ stderr
 backfilling
 timebased
 kedro
+idempotence
  - posts/4-pillars-to-becoming-data-driven.md
 astronomer.io
-_
+default_airflow.cfg file
```
kentdanas, line 960:
```suggestion
default_airflow.cfg
```

PR #209, opened 2021-05-24
Add new testing Airflow guide

kentdanas:
@jwitz new guide on testing Airflow DAGs, would love your review when you have time.
@vparekh94 let me know if you have any thoughts on content.

### DIFF
--- a/guides/testing-airflow.md
+++ b/guides/testing-airflow.md
@@ -8,9 +8,11 @@ tags: ["DAGs", "Best Practices", "Testing"]
 
 ## Overview
 
-One of the core principles of Airflow is that your DAGs are defined as Python code. This allows you to treat your data pipelines like you would any other piece of code and integrate them into a standard software development lifecycle, using source control, CI/CD, and automated testing. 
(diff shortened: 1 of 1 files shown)


jwitz (approved):
Another banger @kentdanas ! Some minor wording and grammar suggestions throughout.

vparekh94:
@kentdanas this is absolutely great! 

Do you think we should link out to a sample repository that shows how some of these tests could be used as a part of CI/CD? I think an example of one of the concepts you shown roped into how this would look with CI/CD with GithubActions could be a great guide for folks to use.

kentdanas:
> @kentdanas this is absolutely great!
> 
> Do you think we should link out to a sample repository that shows how some of these tests could be used as a part of CI/CD? I think an example of one of the concepts you shown roped into how this would look with CI/CD with GithubActions could be a great guide for folks to use.

@vparekh94, yes, great idea! Created a repo here that runs the DAG validation tests in GH actions before (theoretically) deploying to Astronomer https://github.com/astronomer/airflow-testing-guide. Also added a note to the guide linking out to this repo. Let me know what you think!

PR #213, opened 2021-06-04
Add cross-dag dependencies guide

kentdanas:
@jwitz using one of my guide review coupons :)
Would love your thoughts on whether all the explanations are clear enough (especially in the last section), or whether more "something" (screenshots, words, ?) is needed. Didn't want to make it too repetitive, but also want to make sure everything is clear.

### DIFF
--- a/guides/cross-dag-dependencies.md
+++ b/guides/cross-dag-dependencies.md
@@ -1,0 +1,247 @@
+---
+title: "Cross-DAG Dependencies"
+description: "How to implement dependencies between your Airflow DAGs."
+date: 2021-06-07T00:00:00.000Z
+slug: "cross-dag-dependencies"
+tags: ["DAGs", "Subdags"]
+---
+
+## Overview
+
+When designing Airflow DAGs, it is often recommended to put all related tasks in the same DAG. However, sometimes this is not practical, and it is necessary to create dependencies between DAGs. According to the Airflow documentation on [cross-DAG dependencies](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/external_task_sensor.html#cross-dag-dependencies), this can be useful when:
+
+- Two DAGs are dependent but have different schedules
+- Different teams are responsible for the DAGs but there are cross-DAG dependencies
+- A task depends on a task in another DAG, but for a different execution date
+
+For these, or any other scenarios where you have dependent DAGs, we've got you covered! In this guide, we'll discuss multiple methods for implementing cross-DAG dependencies, as well as how to do this if your dependent DAGs are located in different Airflow deployments.
+
+> Note: All code in this guide can be found in [this Github repo](https://github.com/astronomer/cross-dag-dependencies-tutorial).
+
+## Cross-DAG Dependencies
+
+There are multiple ways to implement cross-DAG dependencies in Airflow, including the `TriggerDagRunOperator`, `ExternalTaskSensor`, and the Airflow API. Which you use depends on how your DAG dependencies are configured, and what your Airflow setup looks like. In this section, we detail how to use each method and ideal scenarios for each, as well as how to view dependencies in the Airflow UI.
+
+Note: it can be tempting to use SubDAGs to handle DAG dependencies, but we highly recommend against doing so as SubDAGs can create performance issues. Instead, use one of the other methods described below.
+
+### TriggerDagRunOperator
+
+The `TriggerDagRunOperator` is an easy way to implement cross-DAG dependencies. This operator allows you to have a task in one DAG that triggers another DAG in the same Airflow environment. The Astronomer Registry has more documentation for this operator [here](https://registry.astronomer.io/providers/apache-airflow/modules/triggerdagrunoperator).
+
+The `TriggerDagRunOperator` is ideal in situations where you have one parent DAG, that needs to trigger one or more downstream child DAGs, or if you have dependent DAGs that have both upstream and downstream tasks in the parent DAG. Because you can use this operator for any task in your DAG, it is highly flexible; it can also often be an ideal replacement for SubDAGs. 
+
+Below is an example DAG that implements the `TriggerDagRunOperator` to trigger the `dependent-dag` between two other tasks. 
+
+```python
+from airflow import DAG
+from airflow.operators.python import PythonOperator
+from airflow.operators.trigger_dagrun import TriggerDagRunOperator
+from datetime import datetime, timedelta
+
+def print_task_type(**kwargs):
+    """
+    Dummy function to call before and after depdendent DAG.
+    """
+    print(f"The {kwargs['task_type']} task has completed.")
+
+# Default settings applied to all tasks
+default_args = {
+    'owner': 'airflow',
+    'depends_on_past': False,
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'retries': 1,
+    'retry_delay': timedelta(minutes=5)
+}
+
+with DAG('trigger-dagrun-dag',
+         start_date=datetime(2021, 1, 1),
+         max_active_runs=1,
+         schedule_interval='@daily',
+         default_args=default_args,
+         catchup=False
+         ) as dag:
+
+    start_task = PythonOperator(
+        task_id='starting_task',
+        python_callable=print_task_type,
+        op_kwargs={'task_type': 'starting'}
+    )
+
+    trigger_dependent_dag = TriggerDagRunOperator(
+        task_id="trigger_dependent_dag",
+        trigger_dag_id="dependent-dag",
+        wait_for_completion=True
+    )
+
+    end_task = PythonOperator(
+        task_id='end_task',
+        python_callable=print_task_type,
+        op_kwargs={'task_type': 'ending'}
+    )
+
+    start_task >> trigger_dependent_dag >> end_task
+```
+
+There are a couple of things to note when using this operator:
+
+- If your dependent DAG requires a config input, or a specific execution date, these can be specified in the operator using the `conf` and `execution_date` params respectively.
+- If your parent DAG has downstream tasks that require the child DAG to finish first, you should set the `wait_for_completion` param to `True` as shown in the example above. This param defaults to `False`, meaning once the child DAG has started, the parent DAG will mark the task as a success and move on to any downstream tasks.
+
+### ExternalTaskSensor
+
+The next method for creating cross-DAG dependencies is using the `ExternalTaskSensor`. This method uses a sensor in the child DAG, which will wait until a task is completed in the parent DAG before moving on to other tasks. You can find more info on this sensor [here](https://registry.astronomer.io/providers/apache-airflow/modules/externaltasksensor).
+
+This method is not as flexible as the `TriggerDagRunOperator`, since the dependency is implemented in the downstream DAG. It is ideal in situations where you have a DAG that can only run after another task or DAG(s) has completed. An example DAG using the `ExternalTaskSensor` is shown below:
+
+```python
+from airflow import DAG
+from airflow.operators.python import PythonOperator
+from airflow.sensors.external_task import ExternalTaskSensor
+from datetime import datetime, timedelta
+
+def downstream_fuction():
+    """
+    Downstream function with print statement.
+    """
+    print('Upstream DAG has completed. Starting other tasks.')
+
+default_args = {
+    'owner': 'airflow',
+    'depends_on_past': False,
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'retries': 1,
+    'retry_delay': timedelta(minutes=5)
+}
+
+with DAG('external-task-sensor-dag',
+         start_date=datetime(2021, 1, 1),
+         max_active_runs=3,
+         schedule_interval='*/1 * * * *',
+         catchup=False
+         ) as dag:
+
+    child_task1 = ExternalTaskSensor(
+        task_id="child_task1",
+        external_dag_id='example_dag',
+        external_task_id='bash_print_date2',
+        allowed_states=['success'],
+        failed_states=['failed', 'skipped']
+    )
+
+    child_task2 = PythonOperator(
+        task_id='child_task2',
+        python_callable=downstream_fuction,
+        provide_context=True
+    )
+
+    child_task1 >> child_task2
+```
+
+In this DAG, `child_task1` waits for the `bash_print_date2` task of `example_dag` to complete before moving on to execute the rest of the downstream tasks (`child_task2`). If you want the child DAG to wait for the entire parent DAG to finish instead of a specific task, you can set the `external_task_id` to `None`. Note that in this case we specify that the external task must have a state of `success` in order for the child task to succeed, as defined by the `allowed_states` and `failed_states`. 
+
+Also note that in the example above, the parent DAG (`example_dag`) and child DAG (`external-task-sensor-dag`) must have the same start date and schedule interval. This is because the `ExternalTaskSensor` will look for completion of the specified task or DAG at the same `execution_date`. To look for completion of the external task at a different date, you can make use of either of the `execution_delta` or `execution_date_fn` parameters (these are described in more detail in the documentation linked above).
+
+### Airflow API
+
+The [Airflow API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html) is another way of creating cross-DAG dependencies. This is especially useful in [Airflow 2.0](https://www.astronomer.io/blog/introducing-airflow-2-0), which has a full stable REST API. To use the API to trigger a DAG run, you can make a POST request to the `DAGRuns` endpoint as described in the Airflow documentation [here](https://www.astronomer.io/blog/introducing-airflow-2-0).
+
+This method is useful if your dependent DAGs live in different Airflow environments (more on this below), or if the child DAG does not have any downstream dependencies in the parent DAG. One drawback to this method is that the task triggering the child DAG will complete once the API call is complete, rather than when the child DAG is complete.
+
+Using the API to trigger a child DAG can be implemented within a DAG by using the `SimpleHttpOperator` as shown in the example DAG below:
+
+```python
+from airflow import DAG
+from airflow.operators.python import PythonOperator
+from airflow.providers.http.operators.http import SimpleHttpOperator
+from datetime import datetime, timedelta
+import json
+
+# Define body of POST request for the API call to trigger another DAG
+date = '{{ execution_date }}'
+request_body = {
+  "execution_date": date
+}
+json_body = json.dumps(request_body)
+
+def print_task_type(**kwargs):
+    """
+    Dummy function to call before and after child DAG.
+    """
+    print(f"The {kwargs['task_type']} task has completed.")
+    print(request_body)
+
+default_args = {
+    'owner': 'airflow',
+    'depends_on_past': False,
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'retries': 1,
+    'retry_delay': timedelta(minutes=5)
+}
+
+with DAG('api-dag',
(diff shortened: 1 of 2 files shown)


guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,247 @@
...
+        python_callable=print_task_type,
+        op_kwargs={'task_type': 'ending'}
+    )
+
+    start_task >> trigger_dependent_dag >> end_task
+```
+
+There are a couple of things to note when using this operator:
+
+- If your dependent DAG requires a config input, or a specific execution date, these can be specified in the operator using the `conf` and `execution_date` params respectively.
```
jwitz, line 88:
```suggestion
- If your dependent DAG requires a config input or a specific execution date, these can be specified in the operator using the `conf` and `execution_date` params respectively.
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,247 @@
...
+- If your parent DAG has downstream tasks that require the child DAG to finish first, you should set the `wait_for_completion` param to `True` as shown in the example above. This param defaults to `False`, meaning once the child DAG has started, the parent DAG will mark the task as a success and move on to any downstream tasks.
+
+### ExternalTaskSensor
+
+The next method for creating cross-DAG dependencies is using the `ExternalTaskSensor`. This method uses a sensor in the child DAG, which will wait until a task is completed in the parent DAG before moving on to other tasks. You can find more info on this sensor [here](https://registry.astronomer.io/providers/apache-airflow/modules/externaltasksensor).
```
jwitz, line 93:
```suggestion
The next method for creating cross-DAG dependencies is to add a `ExternalTaskSensor` to your child DAG. The child DAG will wait until a task is completed in the parent DAG before moving on to other tasks. You can find more info on this sensor on the [Astronomer Registry](https://registry.astronomer.io/providers/apache-airflow/modules/externaltasksensor).
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,247 @@
...
+
+This method is not as flexible as the `TriggerDagRunOperator`, since the dependency is implemented in the downstream DAG. It is ideal in situations where you have a DAG that can only run after another task or DAG(s) has completed. An example DAG using the `ExternalTaskSensor` is shown below:
```
jwitz, line 95:
```suggestion
This method is not as flexible as the `TriggerDagRunOperator`, since the dependency is implemented in the downstream DAG. It is ideal in situations where you have a child DAG that can run only after another task in a parent DAG has completed. An example DAG using the `ExternalTaskSensor` is shown below:
```
Rewriting this based on what I assume is going on here, but not super sure especially since we're no longer using the parent/child language.

jwitz (changes requested):
@kentdanas Started suggesting inline changes here, but I hit a wall with understanding what's going on at a high level with dependencies. Before continuing my review, I think this guide could benefit from a few inclusions:

* In the intro, we should define what "parent/child" is. Is it different from "upstream/downstream", and how so? Is it a dependency in and of itself, or is the dependency a separate relationship? Once you get that down, it'll be easier to explain each individual dependency scenario. 
* Even though it's only available in 2.1, could we show each of these methods as they appear in the DAG Dependencies view? We can mention at the beginning that the feature's only available in 2.1 and describe how to access it in the same place as we do now, but having that native Airflow visualization would be super helpful for understanding these examples. 

I'll take another look after you look through these current comments. Excited get this published!

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,247 @@
...
+tags: ["DAGs", "Subdags"]
+---
+
+## Overview
+
+When designing Airflow DAGs, it is often recommended to put all related tasks in the same DAG. However, sometimes this is not practical, and it is necessary to create dependencies between DAGs. According to the Airflow documentation on [cross-DAG dependencies](https://airflow.apache.org/docs/apache-airflow/stable/howto/operator/external_task_sensor.html#cross-dag-dependencies), this can be useful when:
+
+- Two DAGs are dependent but have different schedules
+- Different teams are responsible for the DAGs but there are cross-DAG dependencies
+- A task depends on a task in another DAG, but for a different execution date
```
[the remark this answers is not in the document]
kentdanas: I think it's worded poorly in the Apache doc - they're trying to say that even though the two tasks might be in the same DAG, one task depends on the other for a different execution date, which can't be accomplished in the same DAG Run. So essentially you would have a DAG that has a dependency with itself. That seemed unnecessarily confusing to me, so I simplified a bit here.

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,247 @@
...
+
+Note: it can be tempting to use SubDAGs to handle DAG dependencies, but we highly recommend against doing so as SubDAGs can create performance issues. Instead, use one of the other methods described below.
+
+### TriggerDagRunOperator
+
+The `TriggerDagRunOperator` is an easy way to implement cross-DAG dependencies. This operator allows you to have a task in one DAG that triggers another DAG in the same Airflow environment. The Astronomer Registry has more documentation for this operator [here](https://registry.astronomer.io/providers/apache-airflow/modules/triggerdagrunoperator).
+
+The `TriggerDagRunOperator` is ideal in situations where you have one parent DAG, that needs to trigger one or more downstream child DAGs, or if you have dependent DAGs that have both upstream and downstream tasks in the parent DAG. Because you can use this operator for any task in your DAG, it is highly flexible; it can also often be an ideal replacement for SubDAGs.
```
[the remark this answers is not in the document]
kentdanas: Went through and made the metaphors consistent, and also added some graph view screenshots for the example DAGs. Hopefully it's more clear now.

kentdanas:
@jwitz ok, I updated the language throughout to only use upstream/downstream (I liked this better than parent/child), and re-worked the intro to more thoroughly describe what those terms mean. Also added some clarifying language and screenshots of the example DAGs in the implementation section. Let me know if this helps make it clearer :)

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+title: "Cross-DAG Dependencies"
+description: "How to implement dependencies between your Airflow DAGs."
+date: 2021-06-07T00:00:00.000Z
+slug: "cross-dag-dependencies"
+tags: ["DAGs", "Subdags"]
+---
+
+## Overview
+
+When designing Airflow DAGs, it is often best practice to put all related tasks in the same DAG. However, it's sometimes necessary to create dependencies between your DAGs. One way of thinking about this is that a node in your DAG is actually a whole other DAG, rather than just a single task. Throughout this guide, we'll use "upstream DAG" to denote the DAG (or partial DAG) that must run before the dependent DAG, and the "downstream DAG" to denote the DAG that depends on the upstream DAG's completion before running.
```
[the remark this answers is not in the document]
jwitz: Also...! do we need to more explicitly define "dependent DAG"? My impression is that both upstream and downstream DAGs are dependent DAGs, but it seems like it's mentioned only in reference to downstream DAGs.

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+- If your upstream DAG has downstream tasks that require the downstream DAG to finish first, you should set the `wait_for_completion` param to `True` as shown in the example above. This param defaults to `False`, meaning once the downstream DAG has started, the upstream DAG will mark the task as a success and move on to any downstream tasks.
+
+### ExternalTaskSensor
+
+The next method for creating cross-DAG dependencies is to add a `ExternalTaskSensor` to your downstream DAG. The downstream DAG will wait until a task is completed in the upstream DAG before moving on to the rest of the DAG. You can find more info on this sensor on the [Astronomer Registry](https://registry.astronomer.io/providers/apache-airflow/modules/externaltasksensor).
+
+This method is not as flexible as the `TriggerDagRunOperator`, since the dependency is implemented in the downstream DAG. It is ideal in situations where you have a downstream DAG that can run only after another task in a upstream DAG has completed. An example DAG using the `ExternalTaskSensor` is shown below:
```
jwitz, line 102:
```suggestion
This method is not as flexible as the `TriggerDagRunOperator`, since the dependency is implemented in the downstream DAG. It is ideal in situations where you have a downstream DAG that can run only after a task in the upstream DAG has completed. An example DAG using the `ExternalTaskSensor` is shown below:
```

The "ideal situation" here seems to be reiterating the definition of the TaskSensor itself, is there anything else we can say about the context here that will give people a sense of whether to use this over the `TriggerDagRunOperator`?

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+        python_callable=downstream_fuction,
+        provide_context=True
+    )
+
+    downstream_task1 >> downstream_task2
+```
+
+In this DAG, `downstream_task1` waits for the `bash_print_date2` task of `example_dag` to complete before moving on to execute the rest of the downstream tasks (`downstream_task2`). The graph view of the DAG looks like this:
+
+![External Task Sensor Grapg](https://assets2.astronomer.io/main/guides/cross-dag-dependencies/external_task_sensor_graph.png)
```
jwitz, line 151: Names in this screenshot are inconsistent with the names in the sample DAG.

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+
+If you want the downstream DAG to wait for the entire upstream DAG to finish instead of a specific task, you can set the `external_task_id` to `None`. Note that in this case we specify that the external task must have a state of `success` in order for the downstream task to succeed, as defined by the `allowed_states` and `failed_states`.
```
jwitz, line 153:
```suggestion
If you want the downstream DAG to wait for the entire upstream DAG to finish instead of a specific task, you can set the `external_task_id` to `None`. In this case, we specify that the external task must have a state of `success` for the downstream task to succeed, as defined by the `allowed_states` and `failed_states`. 
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+
+Also note that in the example above, the upstream DAG (`example_dag`) and downstream DAG (`external-task-sensor-dag`) must have the same start date and schedule interval. This is because the `ExternalTaskSensor` will look for completion of the specified task or DAG at the same `execution_date`. To look for completion of the external task at a different date, you can make use of either of the `execution_delta` or `execution_date_fn` parameters (these are described in more detail in the documentation linked above).
+
+### Airflow API
+
+The [Airflow API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html) is another way of creating cross-DAG dependencies. This is especially useful in [Airflow 2.0](https://www.astronomer.io/blog/introducing-airflow-2-0), which has a full stable REST API. To use the API to trigger a DAG run, you can make a POST request to the `DAGRuns` endpoint as described in the Airflow documentation [here](https://www.astronomer.io/blog/introducing-airflow-2-0).
```
jwitz, line 159:
```suggestion
The [Airflow API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html) is another way of creating cross-DAG dependencies. This is especially useful in [Airflow 2.0](https://www.astronomer.io/blog/introducing-airflow-2-0), which has a full stable REST API. To use the API to trigger a DAG run, you can make a POST request to the `DAGRuns` endpoint as described in the [Airflow documentation](https://www.astronomer.io/blog/introducing-airflow-2-0).
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+
+This method is useful if your dependent DAGs live in different Airflow environments (more on this below), or if the downstream DAG does not have any downstream dependencies in the upstream DAG (e.g. the downstream DAG is the last task in the upstream DAG). One drawback to this method is that the task triggering the downstream DAG will complete once the API call is complete, rather than when the downstream DAG is complete.
```
jwitz, line 161: For my own knowledge, why is that last sentence a drawback?

jwitz, line 161:
```suggestion
This method is useful if your dependent DAGs live in different Airflow environments (more on this in the Cross-Deployment Dependencies section below), or if the downstream DAG does not have any downstream dependencies in the upstream DAG (e.g. the downstream DAG is the last task in the upstream DAG). One drawback to this method is that the task triggering the downstream DAG will complete once the API call is complete, rather than when the downstream DAG is complete.
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+    start_task >> api_trigger_dependent_dag >> end_task
+```
+
+This DAG has a similar structure to the `TriggerDagRunOperator` DAG above, but instead uses the `SimpleHttpOperator` to trigger the `dependent-dag` using the Airflow API. The graph view looks like this:
+
+![API Graph View](https://assets2.astronomer.io/main/guides/cross-dag-dependencies/api_graph.png)
+
+In order to use the `SimpleHttpOperator` to trigger another DAG, you need to define the following:
+
+- `endpoint`: this should be of the form `'/api/v1/dags/<dag-id>/dagRuns'` where `<dag-id>` is the ID of the DAG you want to trigger.
```
jwitz, line 232:
```suggestion
- `endpoint`: This should be of the form `'/api/v1/dags/<dag-id>/dagRuns'` where `<dag-id>` is the ID of the DAG you want to trigger.
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+- `data`: to trigger a DAG Run using this endpoint you must provide an execution date. In the example above, we use the `execution_date` of the upstream DAG, but this can be any date of your choosing. You can also specify other information about the DAG run as described in the API documentation linked above.
```
jwitz, line 233:
```suggestion
- `data`: To trigger a DAG Run using this endpoint, you must provide an execution date. In the example above, we use the `execution_date` of the upstream DAG, but this can be any date of your choosing. You can also specify other information about the DAG run as described in the API documentation linked above.
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+- `http_conn_id`: this should be an Airflow connection of type HTTP, with your Airflow domain as the Host, and any authentication provided either as a Login/Password (if using Basic auth), or as a json-formatted Extra. In the example below we use an authorization token.
```
jwitz, line 234:
```suggestion
- `http_conn_id`: This should be an Airflow connection of type HTTP, with your Airflow domain as the Host. Any authentication should be provided either as a Login/Password (if using Basic auth) or as a JSON-formatted Extra. In the example below, we use an authorization token.
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,261 @@
...
+- Using an `ExternalTaskSensor`
+
+Dependencies can be viewed in the UI by going to `Browse` → `DAG Dependencies`. The screenshot below shows the dependencies created by the `TriggerDagRunOperator` and `ExternalTaskSensor` example DAGs in the sections above.
+
+![DAG Dependencies View](https://assets2.astronomer.io/main/guides/cross-dag-dependencies/dag_dependencies_view.png)
+
+
+## Cross-Deployment Dependencies
+
+Sometimes it may be necessary to implement cross-DAG dependencies where the DAGs do not exist in the same Airflow deployment. The `TriggerDagRunOperator` and `ExternalTaskSensor` methods described above are designed to work with DAGs in the same Airflow environment, so are not ideal for cross-Airflow deployments. The Airflow API on the other hand, is perfect for this use case. In this section we'll focus on how to implement this method on the Astronomer platform, but the general concepts will likely be similar wherever your Airflow environments are deployed.
```
jwitz, line 252:
```suggestion
Sometimes it may be necessary to implement cross-DAG dependencies where the DAGs do not exist in the same Airflow deployment. The `TriggerDagRunOperator` and `ExternalTaskSensor` methods described above are designed to work with DAGs in the same Airflow environment, so they are not ideal for cross-Airflow deployments. The Airflow API, on the other hand, is perfect for this use case. In this section, we'll focus on how to implement this method on the Astronomer platform, but the general concepts will likely be similar wherever your Airflow environments are deployed.
```

jwitz (commented):
This is getting close @kentdanas ! Great job redefining some of the terms here. I left a few more comments and questions about the intro and the screen shots, but the rest of the guide looks great.

guides/cross-dag-dependencies.md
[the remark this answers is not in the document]
kentdanas: Both the "upstream" and the "downstream" DAG would be considered dependent DAGs (they are dependent on each other). This isn't any different than task dependencies, which are well known to any airflow user, so I don't really think it's necessary to define it explicitly.

guides/cross-dag-dependencies.md
kentdanas: If you have a task A that has a downstream dependency (more tasks that run after it), you generally want to wait until that task A is completed before moving on the downstream tasks. With the API in this example, your task A is an entire DAG, and taks A will be considered complete once the API call is made, not when the DAG that got triggered finishes. So if your "downstream" DAG takes 3 minutes to complete, it will still be running even though task A is done, which may impact other downstream tasks.

guides/cross-dag-dependencies.md
kentdanas: Yep, good call. Changed to an example that wouldn't work with the `TriggerDagRunOperator`.

guides/cross-dag-dependencies.md
kentdanas: Good catch, I updated the screenshot.

kentdanas:
@jwitz thanks again for your review! I think I addressed all your remaining comments, let me know if you see anything else!

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,264 @@
...
+![Trigger DAG Graph](https://assets2.astronomer.io/main/guides/cross-dag-dependencies/trigger_dag_run_graph.png)
+
+There are a couple of things to note when using this operator:
+
+- If your dependent DAG requires a config input or a specific execution date, these can be specified in the operator using the `conf` and `execution_date` params respectively.
+- If your upstream DAG has downstream tasks that require the downstream DAG to finish first, you should set the `wait_for_completion` param to `True` as shown in the example above. This param defaults to `False`, meaning once the downstream DAG has started, the upstream DAG will mark the task as a success and move on to any downstream tasks.
+
+### ExternalTaskSensor
+
+The next method for creating cross-DAG dependencies is to add a `ExternalTaskSensor` to your downstream DAG. The downstream DAG will wait until a task is completed in the upstream DAG before moving on to the rest of the DAG. You can find more info on this sensor on the [Astronomer Registry](https://registry.astronomer.io/providers/apache-airflow/modules/externaltasksensor).
```
jwitz, line 103:
```suggestion
The next method for creating cross-DAG dependencies is to add an `ExternalTaskSensor` to your downstream DAG. The downstream DAG will wait until a task is completed in the upstream DAG before moving on to the rest of the DAG. You can find more info on this sensor on the [Astronomer Registry](https://registry.astronomer.io/providers/apache-airflow/modules/externaltasksensor).
```

guides/cross-dag-dependencies.md
```diff
@@ -0,0 +1,264 @@
...
+
+```python
+from airflow import DAG
+from airflow.operators.python import PythonOperator
+from airflow.operators.trigger_dagrun import TriggerDagRunOperator
+from datetime import datetime, timedelta
+
+def print_task_type(**kwargs):
+    """
+    Dummy function to call before and after depdendent DAG.
```
jwitz, line 49:
```suggestion
    Dummy function to call before and after dependent DAG.
```

jwitz (approved):
Some minor fixes suggested, but otherwise looks good to go! 😎

PR #214, opened 2021-06-07
Clarifications in best practices guide

kentdanas:
Clarify top-level code section in best practices guide with better language and examples to solve issue here: https://github.com/astronomer/docs/issues/381.


kentdanas:
@wolfier added another example like you suggested and clarified the language a bit. Let me know if you think this is better.

wolfier:
I think I would also mention the use of XCom as well as Airflow Variable would be a good addition!

Otherwise, it looks good to me!

wolfier (approved):

PR #230, opened 2021-08-12
Fix trigger rule formatting

### LINKED ISSUE #229: Formatting error for trigger rules
On the [Managing Dependencies](https://www.astronomer.io/guides/managing-dependencies) page under Trigger Rules, the formatting is wrong for `all_done`.

I think it's caused by a difference in the markdown renderer used on the actual site and GitHub Markdown, because it renders correctly on GitHub. I think in the line above, `upstream_failed` should be `upstream\_failed`

collinmcnulty:
closes #229 

Can't tell if this will fix the issue because I don't know what renderer is used, but this doesn't make anything worse in GFM and it *should* work.


kentdanas:
Looks good to me, I think you're right that this _should_ fix it

kentdanas (approved):

PR #231, opened 2021-08-17
dbt Airflow Guide

denimalpaca:
Create a guide that summarizes our blog posts with Updater on using dbt with Airflow.

- [ ] Add Hero image to S3 bucket


kentdanas:
Hey @denimalpaca thanks for putting this together! I can definitely go through and edit this for grammar and stuff when ready, but to start, a couple of comments on content:

- I think it would be really helpful here to have a high level summary of why you would want to use dbt with Airflow, as opposed to just dbt cloud or something like that. It can be short and part of the Overview section, just a sentence or two, but I think that would help set the stage. 
- I also think it would be helpful to start with a summary of the different methods of using Airflow and dbt together. One thing I don't get from this right now (or from just reading the blog posts separately) is how it all fits together. If I were reading this trying to figure out how to use dbt with Airflow, I would want something upfront that helped me know which section to look at based on my use case (or at least gave me a sense of what was coming throughout the guide). I tend to think of the guides following a track of "why would I want to do this" -> "what are my options for doing this" -> "how can I implement those options"

Happy to chat through this more next week if that would be helpful!

josh-fell:
I flippin' love this! 

Two comments: 
- Is there any way to link to the certified, dbt DAGs on the Registry? The dbt DAGs from the guides are on the Registry and it would be killer to link to them directly where it makes sense. https://registry.astronomer.io/dags?query=dbt&badges=certified

- There have been some standard updates to OSS example DAGs recently which we should start implementing as well: 
  - Almost all of the OSS example DAGs are being updated to remove the "old" structure of having a separate `default_args` dict for the DAG in deference to placing the dict in the `DAG` definition.  This pattern reads better and is more "Pythonic" since the `default_args` variable is only used once so a variable is really a moot approach.  (The only examples that will still have this separate dict is where there are multiple example DAGs in the same file.)
  - Also, if the `default_args` don't _really_ apply to the DAG or are the default values we should remove them. There is a **LOT** of copy pasta out there in the wild so our examples should be streamlined a bit and succinct.
  - Finally, `start_date` should not be part of `default_args`. There is even a CI check in OSS that prevents this behavior in example DAGs.

It would be great to update the DAG definitions to something like this:
```python
dag = DAG(
    'dbt_dag',
    'start_date': datetime(2020, 12, 23),
    default_args={
        'owner': 'astronomer',
        'email_on_failure': False,
        'email_on_retry': False,
        'retries': 1,
    },
    description='An Airflow DAG to invoke simple dbt commands',
    schedule_interval=timedelta(days=1),
    catchup=False,
)
```

PR #232, opened 2021-08-17
Updated Airflow guides based on ebook publication

jwitz:
We updated some existing info in Airflow guides for our upcoming ebook. This PR applies those changes to our published guides. Only changes from the ebook are formatting related.

### DIFF
--- a/guides/from-operators-to-dagruns.md
+++ b/guides/from-operators-to-dagruns.md
@@ -7,49 +7,50 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
 tags: ["Hooks", "Operators", "Tasks", "DAGs"]
 ---
 <!-- markdownlint-disable-file -->
-## How Work Gets Executed
+## How Work Gets Executed in Airflow
 
-**Operators become Tasks**
-- Operators contain the logic - but once they are added to a DAG file with a `task_id` and `dag`, they become **tasks**.
-- Being explicit, when an operator class is instantiated with a `task_id` and `dag` (along with its other settings) it becomes a **task** within a **DAG**.
+### Operators are wrappers around types of work.
 
-**Tasks become Task Instances**
-- Once a series of tasks becomes bundled to the same **DAG** object, the **DAG** can be executed based on its schedule.
-- The scheduler "taps" the **DAG** and begins to execute the **tasks** depending on their dependencies.
-- **Tasks** that get executed have a `execution_date` and are now called **task instances**. These get logged in the metadata database.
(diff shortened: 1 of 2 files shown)


guides/what-is-a-hook.md
```diff
@@ -11,7 +11,7 @@ tags: ["Hooks", "Operators", "Tasks", "Basics"]
 
 Operators are the main building blocks of Airflow, but operators rely heavily upon Hooks to interact with all of their source and destination systems.
 
-Hooks are used as a away to abstract the methods you would use against a source system. Hooks should be used when interacting with any external system.
+Hooks are used as a way to abstract the methods you would use against a source system, but do not contain the logic for how that system is interacted with. Hooks should always be used to connect with any external system.
```
kentdanas, line 14:
```suggestion
Hooks are used as a way to abstract the methods you would use against a source system, but often contain limited logic for how that system is interacted with. Hooks should always be used to connect with any external system.
```

@jwitz reworded this a bit because it's not _entirely_ correct as is. Hooks sometimes contain functionality for actually "doing things" in addition to just connecting to the source system. For example, the S3Hook has methods to load files to S3, delete files, list keys, etc.

guides/from-operators-to-dagruns.md
```diff
@@ -7,49 +7,50 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
-**Success (Green)**: The task executed successfully. <br>
-**Skipped (Pink)**: The task has been skipped due to an upstream condition.<br>
-**Shutdown (Blue)**: The task is up for retry. <br>
-**Removed (Light Grey)**: The task has been removed. <br>
-**Retry (Gold)**: The task is up for retry. <br>
-**Upstream Failed (Orange)**: The task will not be run because of a failed upstream dependency.<br>
-
-![title](https://assets2.astronomer.io/main/guides/states.png)
\ No newline at end of file
+One of the key pieces of data stored in Airflow’s metadata database is **State**. States are used to keep track of how task instances and DAG Runs are performing. In the screenshot below, we can see how states are represented in the Airflow UI:
```
kentdanas, line 32:
```suggestion
One of the key pieces of data stored in Airflow’s metadata database is **State**. States are used to keep track of what condition task instances and DAG Runs are in. In the screenshot below, we can see how states are represented in the Airflow UI:
```
Let me know what you think about this re-word. "Performing" to me indicates it's related to time/memory/cpu usage and is something that can be "tuned". Tried to make it consistent terminology with general definition of a "state"

kentdanas (changes requested):

jwitz:
@kentdanas Thank you for the feedback, I implemented these changes exactly as you wrote them. I'll try to also get these changes in to the ebook if it's not too late!

kentdanas (approved):

PR #234, opened 2021-08-18
Correct dynamically generated dags page wrt scheduler heartbeat

### LINKED ISSUE #233: Dynamically Generated dags execute on min_file_process_interval, not scheduler heartbeat
I believe that dag files are processed every [min_file_process_interval](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#min-file-process-interval) not every [scheduler heartbeat](https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#scheduler-heartbeat-sec). By default, there is a 30x difference in these times (30s vs 1 s).

Our guide on [dynamically generating dags](https://www.astronomer.io/guides/dynamically-generating-dags) repeatedly says it's the scheduler heartbeat, but I believe this is inaccurate.

collinmcnulty:
Closes #233


guides/dynamically-generating-dags.md
```diff
@@ -359,7 +359,8 @@ dag_factory.generate_dags(globals())
...
 
-- Any code in the `DAG_FOLDER` will be executed on every Scheduler heartbeat. Methods where that code is dynamically generating DAGs, such as the single-file method, are more likely to cause performance issues at scale.
-- If the DAG parsing time (i.e. the time to parse all code in the `DAG_FOLDER`) is greater than the Scheduler heartbeat interval, the scheduler can get locked up and tasks won't be executed. If you are dynamically generating DAGs and tasks aren't running, this is a good metric to review to start troubleshooting. 
+- Any code in the `DAG_FOLDER` will be executed either every `min_file_processing_interval` or as fast as the dag file processor can, whichever is less frequent. Methods where that code is dynamically generating DAGs, such as the single-file method, are more likely to cause performance issues at scale.
+- If you are reaching out to a database to create your dags, including Variables from the metadata database, this means you will be querying quite frequently. Be conscious of your databases ability to handle such frequent connections and any costs you may incur per request from your data provider.
```
kentdanas, line 363:
```suggestion
- If you are reaching out to a database to create your DAGs (e.g. grabbing Variables from the metadata database), this means you will be querying quite frequently. Be conscious of your database's ability to handle such frequent connections and any costs you may incur per request from your data provider.
```

guides/dynamically-generating-dags.md
```diff
@@ -359,7 +359,8 @@ dag_factory.generate_dags(globals())
...
+- You can increase `min_file_processing_interval` to a much higher value if you know that your dags are not changing frequently and if you can tolerate some delay in the dynamic dags changing in response to the external source that generates them.
```
kentdanas, line 364:
```suggestion
- To help with potential performance issues, you can increase the `min_file_processing_interval` to a much higher value. Consider this option if you know that your DAGs are not changing frequently and if you can tolerate some delay in the dynamic DAGs changing in response to the external source that generates them.
```

kentdanas (approved):
@collinmcnulty thanks for catching this and making the changes! I made a couple of minor wording updates - if you're okay with those suggestions you can accept them and go ahead and merge.

collinmcnulty:
Thanks @kentdanas, I like your wording updates. I also included a new caveat that we just found on CRE. It applies only to Astronomer, not airflow in general, so I'm not sure exactly how best to include it.

guides/dynamically-generating-dags.md
```diff
@@ -322,7 +322,7 @@ for filename in os.listdir(config_filepath):
 
 ```
 
-Now to generate our DAG files, we can either run this script ad-hoc, as part of our CI/CD workflow, or we could create another DAG that would run it periodically. After running the script, our final directory would look like the example below, where the `include/` directory contains the files shown above, and the `dags/` directory contains the two dynamically generated DAGs:
+Now to generate our DAG files, we can either run this script ad-hoc, as part of our CI/CD workflow, or we could create another DAG that would run it periodically. (If you're using Astronomer Enterprise or Astronomer Cloud, you will need to use the NFS option for your dags folder and run this script with Kubernetes Pod Operator to launch a pod that has the NFS folder mounted with read/write permissions.) After running the script, our final directory would look like the example below, where the `include/` directory contains the files shown above, and the `dags/` directory contains the two dynamically generated DAGs:
```
kentdanas, line 325: @collinmcnulty I'm not sure I totally understand the addition here. This is only relevant if you were trying to have a DAG that ran the script to generate other DAGs, correct? I'm not sure why that would require using KPO? And NFS is only an option for Enterprise, not for Cloud.

If my assumptions are correct, I almost wonder if it would just be better to take out the note about using another DAG to run the DAG creation script. It definitely complicates things and I think is unlikely to be the route most folks would go for using this method (CI/CD is the only way I've actually heard about customers doing this).

collinmcnulty: You are correct in your assumption that this only applies to dag running a script that writes other dags. The problem in Enterprise is that the dags folder on NFS is mounted as read only, so unless you use KPO to create a pod to mount it separately as R/W will you be able to write to it.

I would be fine removing the statement about running it as a dag. We just had a customer do this and it was a complex ticket, so I want to make sure we don't steer anybody into what is unsuspectingly difficult territory.

kentdanas: Got it, yeah let's just remove it! Your change looks good, feel free to merge!

kentdanas (approved):

PR #235, opened 2021-08-18
Fix bug in dynamic dependencies code

snolan-amount:
The code as given does not match to the DAG graph view shown.


kentdanas:
@snolan-amount thanks for the catch

kentdanas (approved):

PR #236, opened 2021-08-19
Update Readme

kentdanas:
Updated the contributing instructions and removed some info that is no longer relevant. Leaving out the section about deploying guides for now since this should be an automated process but is currently being worked on.


jwitz (approved):
Love it! Thanks for keeping this up-to-date @kentdanas

PR #237, opened 2021-08-20
Old guide cleanup

kentdanas:
Removing out of date Airflow competitor comparisons and out of date Astro vs GCC guide


kentdanas:
Also removing:

- Airflow datastores - all content from this one is covered more clearly and in more depth in other guides.
- External Airflow resources - this will be very difficult to keep up to date. If external resources are useful they should be linked from specific guides.
- Kubectl - out of date, and we shouldn't need a guide specifically for this since k8s documentation covers this well and nothing in the guide is Airflow related. Can link to specific kubectl docs/instructions directly from Astro platform docs if needed.

petedejoy:
Just a note- let's make sure we get redirects in place for all of the guides that we're nuking. @medinab can help with this- all of the routes we pull out should just link back to `/guides`.

medinab:
redirect have been deployed for the site: https://github.com/astronomer/astro-site/pull/448

Merging!

PR #257, opened 2021-08-31
Airflow pools

### LINKED ISSUE #239: Airflow Pools Guide

kentdanas:
@jwitz pretty short and simple one. Feedback appreciated as always!

Closes #239


guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+title: "Airflow Pools"
+description: "Using pools to control task parallelism in Airflow."
+date: 2021-09-01T00:00:00.000Z
+slug: "airflow-pools"
+tags: ["Parallelism", "Tasks"]
+---
+
+## Overview
+
+One of the great benefits to Apache Airflow is that it is built to scale. With the right supporting infrastructure, you can run many tasks in parallel seamlessly. But sometimes that horizontal scalability also necessitates some guardrails: you might have many tasks that interact with the same source system, like an API or database, that you don't want to be overwhelmed with requests. Fortunately, [pools](https://airflow.apache.org/docs/apache-airflow/stable/concepts/pools.html) are an Airflow feature designed for exactly this use case. They allow you to limit parallelism for a chosen set of tasks, giving you fine grained control over when your tasks are run.
```
jwitz, line 11:
```suggestion
One of the great benefits to Apache Airflow is that it is built to scale. With the right supporting infrastructure, you can run many tasks in parallel seamlessly. But sometimes that horizontal scalability also necessitates some guardrails. For example, you might have many tasks that interact with the same source system, like an API or database, that you don't want to be overwhelmed with requests. Fortunately, Airflow's [pools](https://airflow.apache.org/docs/apache-airflow/stable/concepts/pools.html) are designed for exactly this use case. Pools limit parallelism for a chosen set of tasks, giving you fine grained control over when your tasks are run.
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+In this guide, we will cover the basic concepts of Airflow pools including how they are created, how they are assigned, and what you can and can't do with them. We will also provide a couple of examples of implementing pools for different desired scenarios.
```
jwitz, line 13:
```suggestion
In this guide, we will cover the basic concepts of Airflow pools including how to create them, how to assign them, and what you can and can't do with them. We will also walk through some sample DAGs that use pools to fulfill a simple use case. 
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+In this guide, we will cover the basic concepts of Airflow pools including how they are created, how they are assigned, and what you can and can't do with them. We will also provide a couple of examples of implementing pools for different desired scenarios. 
+
+## Airflow Pool Concepts
+
+Airflow pools allow you to limit parallelism for an arbitrary set of tasks. They are often used in cases where you want to limit the number of parallel tasks that *do a certain thing.* For example, tasks that make requests to the same API or database, or tasks that will run on a GPU node of a kubernetes cluster. Below we cover how to create and assign tasks to a pool, and some limitations to be aware of when working with them.
```
jwitz, line 17:
```suggestion
Airflow pools limit parallelism for an arbitrary set of tasks. They are often used in cases where you want to limit the number of parallel tasks that *do a certain thing.* For example, tasks that make requests to the same API or database, or tasks that will run on a GPU node of a Kubernetes cluster. Below we cover how to create and assign tasks to a pool, and some limitations to be aware of when working with them.
```

jwitz: This info seems redundant to the overview, can you:

- Consolidate this info into the overview
- Give the next 3 sections `##` headers

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Creating a Pool
+
+There are three ways you can create and manage a pool in Airflow.
+
+1. Create a pool through the Airflow UI. Go to `Admin` → `Pools` and add a new record. You can define a name, the number of slots, and a description.
```
jwitz, line 23:
```suggestion
- **Create a pool through the Airflow UI:** Go to `Admin` → `Pools` and add a new record. You can define a name, the number of slots, and a description.
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+    ![Pools UI](https://assets2.astronomer.io/main/guides/airflow-pools/pools_ui.png)
+
+2. Create a pool using the Airflow CLI with the `airflow pools` command. The `set` subcommand can be used to create a new pool. Check out the [Airflow CLI documentation](https://airflow.apache.org/docs/apache-airflow/stable/cli-and-env-variables-ref.html#pools) for the full list of pool commands. With the CLI, you can also import pools from a JSON file with the `import` subcommand. This can be useful if you have a large number of pools to define and doing so programmatically would be more efficient.
```
jwitz, line 27:
```suggestion
- **Create a pool using the Airflow CLI with the `airflow pools` command:** The `set` subcommand can be used to create a new pool. Check out the [Airflow CLI documentation](https://airflow.apache.org/docs/apache-airflow/stable/cli-and-env-variables-ref.html#pools) for the full list of pool commands. With the CLI, you can also import pools from a JSON file with the `import` subcommand. This can be useful if you have a large number of pools to define and doing so programmatically would be more efficient.
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+3. Create a pool is using the Airflow REST API. Note that this is only available if you are using Airflow 2. To create a pool, simply submit a POST request with the name and number of slots as the payload. For more information on working with pools from the API, check out the [API documentation](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/post_pool).
```
jwitz, line 28:
```suggestion
- **Create a pool using the Airflow REST API:** Note that this is only available if you are using Airflow 2.0+. To create a pool, simply submit a POST request with the name and number of slots as the payload. For more information on working with pools from the API, check out the [API documentation](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/post_pool).
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+```python
+task_a = PythonOperator(
+        task_id='task_a',
+        python_callable=sleep_function,
+        pool='single_task_pool'
+    )
+```
+
+When tasks are assigned to a pool, they will be scheduled as normal until all the slots are filled. As slots open up, the remaining queued tasks will start running. You can control which tasks within the pool are run first by assigning [priority weights](https://airflow.apache.org/docs/apache-airflow/stable/concepts/priority-weight.html#concepts-priority-weight), which define priorities for the executor queue. These are assigned at the pool level with the `priority_weights` parameter. The values can be any arbitrary integer (default is 1), but higher values get higher priority.
```
jwitz, line 44:
```suggestion
When tasks are assigned to a pool, they will be scheduled as normal until all of the pool's slots are filled. As slots open up, the remaining queued tasks will start running. You can control which tasks within the pool are run first by assigning [priority weights](https://airflow.apache.org/docs/apache-airflow/stable/concepts/priority-weight.html#concepts-priority-weight) which define priorities for the executor queue. These are assigned at the pool level with the `priority_weights` parameter. The values can be any arbitrary integer (default is 1), and higher values get higher priority.
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+For example, in the DAG snippet below `task_a` and `task_b` are both assigned to the `single_task_pool` which has one slot. `task_b` will be executed first because it has a higher priority weight (since we didn't define a priority weight for `task_a`, it gets the default of 1).
```
jwitz, line 46:
```suggestion
For example, in the DAG snippet below `task_a` and `task_b` are both assigned to the `single_task_pool` which has one slot. `task_b` has a priority weight of 2, while `task_a` has the default priority weight of 1. Therefore, `task_b` would be executed first.
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Limitations of Pools
+
+When working with pools, there are a couple of limitations to keep in mind:
+
+- Each task can only be assigned to a single pool.
+- If you are using SubDAGs, pools must be applied directly on tasks inside the SubDAG. Pools set within the `SubDAGOperator` will not be honored.
+- Pools are meant to control parallelism for Task Instances. If instead you are looking to place limits on the number of concurrent DagRuns for a single DAG or all DAGs, check out the `max_active_runs` and `core.max_active_runs_per_dag` parameters respectively.
+
+## Example
```
jwitz, line 80:
```suggestion
## Example: Limit Tasks Hitting an API Endpoint
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+
+Below we show a simple example of how to implement a pool to control the number of tasks hitting an API endpoint. 
+
+In this scenario, we have five tasks across two different DAGs that hit the API and that may run concurrently based on the DAG schedules. But we only want a maximum of three tasks hitting the API at a time, so we create a pool called `api_pool` with three slots. We want the tasks in the `pool_priority_dag` to be prioritized if the pool is filled.
+
+In the `pool_priority_dag` below, all three of the tasks hit the API endpoint and should all be assigned to the pool, so we define the `pool` argument in the DAG `default_args` to apply to all tasks. We also want all three of these tasks to have the same priority weight and for them to be prioritized over tasks in the second DAG, so we assign a `priority_weight` of 3 as a default argument. Note this value is arbitrary: we could have assigned any integer that is higher than the priority weights defined in the second DAG to prioritize these tasks.
```
jwitz, line 86:
```suggestion
In the `pool_priority_dag` below, all three of the tasks hit the API endpoint and should all be assigned to the pool, so we define the `pool` argument in the DAG `default_args` to apply to all tasks. We also want all three of these tasks to have the same priority weight and for them to be prioritized over tasks in the second DAG, so we assign a `priority_weight` of 3 as a default argument. Note that this value is arbitrary: we could have assigned any integer that is higher than the priority weights defined in the second DAG to prioritize these tasks.
```

guides/airflow-pools.md
```diff
@@ -0,0 +1,177 @@
...
+        python_callable=api_function
+    )
+
+    task_c = PythonOperator(
+        task_id='task_c',
+        python_callable=api_function
+    )
+```
+
+Then we have the second DAG, `pool_unimportant_dag` as defined below. In this DAG we have two tasks that hit the API endpoint and should be assigned to the pool, but we also have two other tasks that do not hit the API. Therefore we assign the pool and priority weights within the `PythonOperator` instantiations. We also want to prioritize `task_x` over `task_y` while keeping both at a lower priority than the tasks in the first DAG, so we assign `task_x` a priority weight of 2, and leave `task_y` with the default priority weight of 1.
```
jwitz, line 128:
```suggestion
Then we have the second DAG, `pool_unimportant_dag` as defined below. In this DAG, we have two tasks that hit the API endpoint and should be assigned to the pool, but we also have two other tasks that do not hit the API. Therefore, we assign the pool and priority weights within the `PythonOperator` instantiations. 

We also want to prioritize `task_x` over `task_y` while keeping both at a lower priority than the tasks in the first DAG. To do this we assign `task_x` a priority weight of 2 and leave `task_y` with the default priority weight of 1. 
```

jwitz (approved):
One high level comment about the "Concepts" section but other than that LGTM @kentdanas !

PR #264, opened 2021-09-08
Fix outdated links to incubator-airflow/master branch

BasPH:
This PR started with some updates of broken/outdated links to the old incubator-airflow repo and master branch.

It turned out the markdown-link-check tool fails on filenames with whitespaces so it failed on `example_code/gcp/notebook/Airflow+with+GCP (1).md`. markdown-link-check currently only runs on updated files. We could have reverted the change or renamed the file, but after discussion with @kentdanas decided to remove the complete `example_code` directory since it's outdated/unmaintained anyways.


kentdanas:
@BasPH can you fix the broken link check? Otherwise looks good to me!

kentdanas:
@BasPH looks like the link check is still failing after you bumped the version - any idea what's going on there?

BasPH:
@kentdanas No clue... markdown-link-check is complaining about "ERROR: File not found! Please provide a valid filename as an argument."

I read about a similar bug and figured bumping might fix it, but it doesn't. Enabled verbose mode but it turns out no additional information is printed (looking for a list of files it's going to check). Tried reproducing locally with a Docker image using `ghcr.io/tcort/markdown-link-check:stable`, but was unable to reproduce. Any ideas?

.github/workflows/main.yml
```diff
@@ -31,9 +31,12 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@main
+      - name: Set print commands
+        run: set -x
```
kentdanas, line 35: @BasPH what does this run command do? Trying to look for things that might have changed in this branch that are causing the failure, but everything else looks the same other than the version bump...

BasPH: https://github.com/gaurav-nelson/github-action-markdown-link-check calls a wrapper Bash script which propagates some settings to https://github.com/tcort/markdown-link-check. I figured I could track what's happening by setting `set -x` but it starts in a Docker container, so doesn't work... so reverted the change.

kentdanas (approved):

PR #277, opened 2021-09-20
Error notifications bug fix

### LINKED ISSUE #275: Error Notifications guide mentions non-existent param `on_success_email`
The [Error Notifications in Airflow](https://www.astronomer.io/guides/error-notifications-in-airflow) guide mentions the use of a parameter called `on_success_email` to send an email notification upon a task's successful completion.  Unfortunately this parameter doesn't currently exist.

>Finally, you can also set any task to email on success by setting the email_on_success parameter to True. This is useful when your pipelines have conditional branching, and you want to be notified if a certain path is taken (i.e. certain tasks get run).

kentdanas:
Remove  `email_on_success` reference since it doesn't actually exist, and provide example using `on_success_callback`. 

Closes #275

### DIFF
--- a/guides/error-notifications-in-airflow.md
+++ b/guides/error-notifications-in-airflow.md
@@ -150,7 +150,7 @@ with DAG('sample_dag',
 	)
 ```
 
-Note that custom notification functions can be used in addition to email notifications. If you want to send emails for successful task runs, you can define an email function in your `on_success_callback`, which may look something like this:
+Note that custom notification functions can also be used to send email notifications. For example, if you want to send emails for successful task runs, you can define an email function in your `on_success_callback`, which may look something like this:
 
 ```python
 from airflow.utils.email import send_email
@@ -164,7 +164,7 @@ def success_email_function(context):
 
 ```
 
-This may be useful when your pipelines have conditional branching, and you want to be notified if a certain path is taken (i.e. certain tasks get run).
+This functionality may also be useful when your pipelines have conditional branching, and you want to be notified if a certain path is taken (i.e. certain tasks get run).
 
 ## Email Notifications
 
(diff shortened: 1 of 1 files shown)


guides/error-notifications-in-airflow.md
```diff
@@ -152,7 +150,21 @@ with DAG('sample_dag',
 	)
 ```
 
-Note that custom notification functions can be used in addition to email notifications.
+Note that custom notification functions can be used in addition to email notifications. If you want to send emails for successful task runs, you can define an email function in your `on_success_callback`, which may look something like this:
```
josh-fell, line 153: > Note that custom notification functions can be used in addition to email notifications. If you want to send emails...

Maybe this should read: "Note that custom notification functions can also be used to send email notifications. For example, If you want to send emails..."

guides/error-notifications-in-airflow.md
```diff
@@ -152,7 +150,21 @@ with DAG('sample_dag',
...
+def success_email_function(context):
+    dag_run = context.get('dag_run')
+
+    msg = "DAG ran successfully"
+    subject = f"DAG {dag_run} has completed"
+    send_email(to=your_emails, subject=subject, html_content=msg)
+
+```
+
+This may be useful when your pipelines have conditional branching, and you want to be notified if a certain path is taken (i.e. certain tasks get run).
```
josh-fell, line 167: > This may be useful when your pipelines have conditional branching...

What you think of: "This functionality may also be useful when ..."

guides/error-notifications-in-airflow.md
kentdanas: Agree, will update

guides/error-notifications-in-airflow.md
kentdanas: Yep, updated

PR #292, opened 2021-11-30
dags_folder param cleanup

kentdanas:
@josh-fell `DAG_FOLDER`....couldn't unsee the inconsistency :)


josh-fell (approved):
Probably safe to say we are both afflicted with the same curse. LGTM 👍

PR #306, opened 2021-12-29
Astro for ETL guide

### LINKED ISSUE #294: `astro` for ELT Guide
Guide on using `astro` project functions/decorators for ELT framework, including before/after examples.

kentdanas:
Closes #294 

@jwitz one more for you :) This will be the first of our use case driven guides to promote the `astro` library, showing a before/after ETL implementation.

### DIFF
--- a/guides/astro-etl.md
+++ b/guides/astro-etl.md
@@ -170,7 +170,7 @@ def transform_data(df: pd.DataFrame):
 @dag(start_date=datetime(2021, 12, 1), schedule_interval='@daily', catchup=False)
 
 def astro_billing_dag():
-    # Load subscripton data
+    # Load subscription data
     load_subscription_data = aql.load_file(
         path=S3_FILE_PATH + '/subscription_data.csv',
         file_conn_id="my_s3_conn",
--- a/guides/astro-etl.md
+++ b/guides/astro-etl.md
@@ -9,7 +9,7 @@ tags: ["astro", "ETL", "SQL"]
 
 ## Overview
 
-The `[astro` library](https://github.com/astro-projects/astro) is an open source Python package maintained by Astronomer that provides tools to improve the DAG authoring experience for Airflow users. The available decorators and functions allow you to write DAGs based on how you want your data to move, while simplifying the transformation process between different environments.
+The [`astro` library](https://github.com/astro-projects/astro) is an open source Python package maintained by Astronomer that provides tools to improve the DAG authoring experience for Airflow users. The available decorators and functions allow you to write DAGs based on how you want your data to move, while simplifying the transformation process between different environments.
 
 In this guide, we’ll demonstrate how you can use `astro` functions for ETL use cases. The resulting DAGs will be easier to write and read, and require less code.
 
@@ -25,7 +25,7 @@ More specifically, `astro` has the following functions that are helpful when imp
 
 In the next section, we’ll show a practical example implementing these functions.
 
-## Example Implementation
+## Example ETL Implementation
 
 To show `astro` for ETL in action, we’ll take a pretty common use case: managing billing data. We need to extract customer subscription data by joining data from a CSV on S3 with the results of a query to our Snowflake database. Then we perform some transformations on the data before loading it into a results table. Below we’ll show what the original DAG looked like, and then how `astro` can make it easier.
 
@@ -180,7 +180,7 @@ def astro_billing_dag():
     extracted_data = extract_data(subscriptions=load_subscription_data,
                                     customer_data=Table('customer_data', conn_id=SNOWFLAKE_CONN_ID, schema='SANDBOX_KENTEND'))
 
-    transformed_data = transform_data(extracted_data, output_table=Table('aggregated_bills', conn_id=SNOWFLAKE_CONN_ID))
+    transformed_data = transform_data(extracted_data, output_table=Table('aggregated_bills'     conn_id=SNOWFLAKE_CONN_ID))
     
     # Append transformed data to billing table
     load_transformed_data = aql.append(
@@ -199,4 +199,8 @@ astro_billing_dag = astro_billing_dag()
 
 The key differences in this implementation are:
 
--
+- The `astro` functions `load_file` and `append` take care of loading our raw data from S3 and transformed data to our reporting table respectively. We didn't have to write any explicit code to get the data into Snowflake.
+- Using the `transform` function, we can easily execute SQL to combine our data from multiple tables. The results are automatically stored in a table in Snowflake. We didn't have to use the `SnowflakeHook` or write any of the code to execute the query.
+- We seamlessly transition to a transformation in Python with the `df` function, without needing to explicitly convert the results of our previous task to a Pandas dataframe. We then write the output of our transformation to our "aggregated_bills" table in Snowflake using the `output_table` parameter, so we don't have to worry about storing the data in XCom.
+
+Overall, our DAG with `astro` is shorter, simpler to implement, and easier to read. This allows us to implement even more complicated use cases easily while focusing on the movement of our data.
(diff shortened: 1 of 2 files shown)


guides/astro-etl.md
```diff
@@ -0,0 +1,206 @@
...
+classic_billing_dag = classic_billing_dag()
+```
+
+![Classic Graph](https://assets2.astronomer.io/main/guides/astro-etl/classic_billing_graph_view.png)
+
+While we achieved our ETL goal with the DAG above, there are a couple of limitations that make the implementation more complicated:
+
+- Since there is no way to pass results from `SnowflakeOperator` query to the next task, we have to write our query in a `_DecoratedPythonOperator` function using the `SnowflakeHook`, and do the conversion from SQL to a dataframe explicitly ourselves.
+- Some of our transformations are better suited to SQL, and others are better suited to Python, but transitioning between the two requires extra boilerplate code to explicitly make those conversions.
+- While the TaskFlow API makes it easier to pass data between tasks here, it is storing the resulting dataframes in as XComs by default. This means we need to worry about the size of our data. We could implement a custom XCom backend, but that would be extra lift.
```
jwitz, line 133:
```suggestion
- While the TaskFlow API makes it easier to pass data between tasks here, it is storing the resulting dataframes as XComs by default. This means that we need to worry about the size of our data. We could implement a custom XCom backend, but that would be extra lift.
```

guides/astro-etl.md
```diff
@@ -0,0 +1,206 @@
...
+- Loading data back to Snowflake after the transformation is complete requires storing an intermediate CSV in S3, and writing the code to do so ourselves.
```
jwitz, line 134:
```suggestion
- Loading data back to Snowflake after the transformation is complete requires writing extra code to store an intermediate CSV in S3.
```

guides/astro-etl.md
```diff
@@ -0,0 +1,206 @@
...
+    transformed_data >> load_transformed_data
+
+astro_billing_dag = astro_billing_dag()
+```
+
+![Astro Graph](https://assets2.astronomer.io/main/guides/astro-etl/astro_billing_graph_view.png)
+
+The key differences in this implementation are:
+
+- The `astro` functions `load_file` and `append` take care of loading our raw data from S3 and transformed data to our reporting table respectively. We didn't have to write any explicit code to get the data into Snowflake.
```
jwitz, line 202:
```suggestion
- The `load_file` and `append` functions take care of loading our raw data from S3 and transforming data to our reporting table respectively. We didn't have to write any extra code to get the data into Snowflake.
```

guides/astro-etl.md
```diff
@@ -0,0 +1,206 @@
...
+- Using the `transform` function, we can easily execute SQL to combine our data from multiple tables. The results are automatically stored in a table in Snowflake. We didn't have to use the `SnowflakeHook` or write any of the code to execute the query.
```
jwitz, line 203:
```suggestion
- Using the `transform` function, we easily executed SQL to combine our data from multiple tables. The results are automatically stored in a table in Snowflake. We didn't have to use the `SnowflakeHook` or write any of the code to execute the query.
```

guides/astro-etl.md
```diff
@@ -0,0 +1,206 @@
...
+- We seamlessly transition to a transformation in Python with the `df` function, without needing to explicitly convert the results of our previous task to a Pandas dataframe. We then write the output of our transformation to our "aggregated_bills" table in Snowflake using the `output_table` parameter, so we don't have to worry about storing the data in XCom.
```
jwitz, line 204:
```suggestion
- We seamlessly transitioned to a transformation in Python with the `df` function without needing to explicitly convert the results of our previous task to a Pandas dataframe. We then wrote the output of our transformation to our "aggregated_bills" table in Snowflake using the `output_table` parameter, so we didn't have to worry about storing the data in XCom.
```

jwitz (approved):
Looks great! Just a few small suggestions

guides/astro-etl.md
```diff
@@ -0,0 +1,206 @@
...
+                                    columns=['TYPE'], 
+                                    aggfunc='count').reset_index()
+
+    return transformed_df
+
+
+@dag(start_date=datetime(2021, 12, 1), schedule_interval='@daily', catchup=False)
+
+def astro_billing_dag():
+    # Load subscripton data
```
jwitz, line 173:
```suggestion
    # Load subscription data
```

guides/astro-etl.md
```diff
@@ -0,0 +1,211 @@
...
+        extracted_data,
+        output_table=Table('aggregated_bills')
+    )
+    
+    # Append transformed data to billing table
+    append(
+        conn_id="snowflake",
+        append_table=transformed_data,
+        columns=["DATE", "CUSTOMER_ID", "AMOUNT"],
+        main_table=main_table,
```
josh-fell, line 196:
```suggestion
        main_table=Table("billing_reporting", schema="SANDBOX_KENTEND"),
```
IMO this is more readable than declaring a `main_table` variable above since it's only used once.

guides/astro-etl.md
```diff
@@ -0,0 +1,211 @@
...
+
+SNOWFLAKE_CONN_ID = "snowflake"
+S3_FILE_PATH = '</path/to/file/'
+
+# Start by selecting data from two source tables in Snowflake
+@transform
+def extract_data(subscriptions: Table, customer_data: Table):
+    return """SELECT * FROM {subscriptions}
+    LEFT JOIN {customer_data}
+    ON customer_id=customer_id"""
```
josh-fell, line 157:
```suggestion
    return """
    SELECT *
    FROM {subscriptions}
    LEFT JOIN {customer_data}
           ON customer_id=customer_id
    """
```
WDYT about this for readability or something similar? I think the newlines help a little.

josh-fell (commented):
Minor comments and musings.

guides/astro-etl.md
```diff
@@ -0,0 +1,211 @@
...
+SNOWFLAKE_CONN_ID = 'snowflake'
+SNOWFLAKE_SCHEMA = 'schema_name'
+SNOWFLAKE_STAGE = 'stage_name'
+SNOWFLAKE_WAREHOUSE = 'warehouse_name'
+SNOWFLAKE_DATABASE = 'database_name'
+SNOWFLAKE_ROLE = 'role_name'
+SNOWFLAKE_SAMPLE_TABLE = 'sample_table'
+SNOWFLAKE_RESULTS_TABLE = 'result_table'
+
+@task(task_id='extract_data')
```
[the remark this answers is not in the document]
kentdanas: I did want to show that you can pass in a different `task_id`, since it's a question that comes up a lot :)

guides/astro-etl.md
```diff
@@ -0,0 +1,211 @@
...
+
+### The DAG Before Astro
+
+Here is our billing subscription ETL DAG implemented with OSS Airflow operators and decorators, as well as the TaskFlow API:
+
+```python
+from datetime import datetime
+import pandas as pd
+
+from airflow import DAG
```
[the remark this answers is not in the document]
kentdanas: Good call

guides/astro-etl.md
```diff
@@ -0,0 +1,211 @@
...
+        s3_keys=[S3_FILE_PATH + '/trasnformed_file_name.csv'],
+        table=SNOWFLAKE_RESULTS_TABLE,
+        schema=SNOWFLAKE_SCHEMA,
+        stage=SNOWFLAKE_STAGE,
+        file_format="(type = 'CSV',field_delimiter = ',')",
+    )
+
+    extracted_data = extract_data()
+    transformed_data = transform_data(extracted_data)
+    load_subscription_data >> extracted_data >> transformed_data >> load_transformed_data
```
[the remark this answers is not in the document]
kentdanas: Ugh I tried that, but that doesn't maintain the `load_subscription_data >> extracted_data` dependency. I feel like there should be a cleaner way, but I'm not sure what it is.

PR #328, opened 2022-02-10
Pools guide minor note update

kentdanas:
Add note to pools guide on what happens with a nonexistent pool assignment

### DIFF
--- a/guides/airflow-pools.md
+++ b/guides/airflow-pools.md
@@ -39,7 +39,11 @@ task_a = PythonOperator(
     )
 ```
 
-When tasks are assigned to a pool, they will be scheduled as normal until all of the pool's slots are filled. As slots open up, the remaining queued tasks will start running. You can control which tasks within the pool are run first by assigning [priority weights](https://airflow.apache.org/docs/apache-airflow/stable/concepts/priority-weight.html#concepts-priority-weight) which define priorities for the executor queue. These are assigned at the pool level with the `priority_weights` parameter. The values can be any arbitrary integer (default is 1), and higher values get higher priority.
+When tasks are assigned to a pool, they will be scheduled as normal until all of the pool's slots are filled. As slots open up, the remaining queued tasks will start running. 
+
+> **Note:** If you assign a task to a pool that doesn't exist, the task will not be scheduled when the DAG runs. Currently this will not result in an error in the task logs or Airflow UI.
+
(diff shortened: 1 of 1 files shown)


guides/airflow-pools.md
```diff
@@ -39,7 +39,11 @@ task_a = PythonOperator(
     )
 ```
 
-When tasks are assigned to a pool, they will be scheduled as normal until all of the pool's slots are filled. As slots open up, the remaining queued tasks will start running. You can control which tasks within the pool are run first by assigning [priority weights](https://airflow.apache.org/docs/apache-airflow/stable/concepts/priority-weight.html#concepts-priority-weight) which define priorities for the executor queue. These are assigned at the pool level with the `priority_weights` parameter. The values can be any arbitrary integer (default is 1), and higher values get higher priority.
+When tasks are assigned to a pool, they will be scheduled as normal until all of the pool's slots are filled. As slots open up, the remaining queued tasks will start running. 
+
+> **Note:** If you assign a task to a pool that doesn't exist, the task will not be scheduled when the DAG runs. Currently this will not result in an error in the task logs or Airflow UI.
```
jwitz, line 44:
```suggestion
> **Note:** If you assign a task to a pool that doesn't exist, then the task will not be scheduled when the DAG runs. There are currently no errors or checks for this in the Airflow UI, so be sure to double check the name of the pool that you're assigning a task to.
```

guides/airflow-pools.md
```diff
@@ -39,7 +39,11 @@ task_a = PythonOperator(
...
+
+You can control which tasks within the pool are run first by assigning [priority weights](https://airflow.apache.org/docs/apache-airflow/stable/concepts/priority-weight.html#concepts-priority-weight) which define priorities for the executor queue. These are assigned at the pool level with the `priority_weights` parameter. The values can be any arbitrary integer (default is 1), and higher values get higher priority.
```
jwitz, line 46:
```suggestion
You can control which tasks within the pool are run first by assigning [priority weights](https://airflow.apache.org/docs/apache-airflow/stable/concepts/priority-weight.html#concepts-priority-weight). These are assigned at the pool level with the `priority_weights` parameter. The values can be any arbitrary integer (default is 1), and higher values get higher priority.
```

guides/airflow-pools.md
jwitz: Assuming this is why the note exists? Just trying to make the implications of this behavior more clear

jwitz (approved):
Two small comments, otherwise LGTM

guides/airflow-pools.md
kentdanas: Yep, exactly. Thanks for the edit

PR #342, opened 2022-02-22
Update links to new Astro docs

### LINKED ISSUE #335: Update Astronomer docs links in guides
Most are outdated and pointing to our old docs site. Redirects should be in place, but should get them updated at some point (especially references to nebula docs).

kentdanas:
@jwitz went through a did a pass at updating Astronomer documentation links in the guides. Also tagging along a few updates of references to "astronomer platform" that I had missed previously. 

In most cases I defaulted to Astro docs, unless there was only a relevant doc in the Software section, and tried to make sure to update the language to match the use case. Definitely possible we will need more updates in this area as Astro functionality grows, but can tackle that as it comes.

Closes #335


jwitz (changes requested):
Thank you for doing this pass @kentdanas ! It looks like we're defaulting to the Astro docs for CLI installation - in that case, we should also update all references of "Astronomer CLI" to "Astro CLI", since that's now a separate tool/ installation from the OG Astronomer CLI: https://www.notion.so/astronomerio/Astro-Rebrand-Plan-Terminology-f1644928e5224e57b6318332852d1ee3

kentdanas:
@jwitz good catch! Think I have all of those references updated now.

kentdanas:
@jwitz pinging on this in case it got buried. Let me know if you see any other needed updates.

jwitz:
Ah woops, LGTM!

jwitz (approved):

PR #354, opened 2022-04-08
Update imports in airflow-redshift.md

magdagultekin:
`datetime` has been imported twice in two example DAGs.


kentdanas (approved):
Good catch @magdagultekin!

PR #358, opened 2022-04-18
Update idempotency definition

kentdanas:


guides/dag-best-practices.md
```diff
@@ -21,9 +21,9 @@ For an in-depth walk through and examples of some of the concepts covered here,
 
 Before we jump into best practices specific to Airflow, we need to review one concept which applies to all data pipelines. 
 
-[Idempotency](https://en.wikipedia.org/wiki/Idempotence) is the foundation for many computing practices, including the Airflow best practices in this guide. Specifically, it is a quality: A computational operation is considered idempotent if it always produces the same output.
+[Idempotency](https://en.wikipedia.org/wiki/Idempotence) is the foundation for many computing practices, including the Airflow best practices in this guide. Specifically, it is a quality: A program is considered idempotent if running the program on a set input once has the same effect as running the program multiple times.
```
collinmcnulty, line 24: I feel "on a set input" is placed oddly. What do you think of these alternatives?

A program is considered idempotent if running the program on a set input once has the same effect as running the program multiple times on that same input.

A program is considered idempotent if, for a set input, running the program once has the same effect as running the program multiple times.

guides/dag-best-practices.md
```diff
@@ -21,9 +21,9 @@ For an in-depth walk through and examples of some of the concepts covered here,
...
 
-In the context of Airflow, a DAG is considered idempotent if every DAG Run generates the same results even when run multiple times. Designing idempotent DAGs decreases recovery time from failures and prevents data loss.
+In the context of Airflow, a DAG is considered idempotent if rerunning the same *DAG Run* (with the same inputs) multiple times has the same effect as running it only once. Designing idempotent DAGs decreases recovery time from failures and prevents data loss.
```
collinmcnulty, line 26: Idempotence also exists on the task level. Making individual tasks idempotent helps with things like clear-and-rerun. I think the best way to make idempotent DAGs is to make a DAG out of tasks which are all idempotent. Maybe we should call that out?

collinmcnulty (changes requested):

guides/dag-best-practices.md
kentdanas: Updated to the second one.

guides/dag-best-practices.md
kentdanas: Yeah good call, added some language around tasks.

dylanbstorey: Do we have a “Task” best practices document for that, I see your point but it doesn’t feel like a natural aside ( or maybe we could just use an admonition block to make an explicit block out of it as a comment on the larger document ?)

kentdanas: We don't have a separate task best practices guide, but many of the best practices stated later in this guide are meant for tasks (and indicated as such) and are tied back to idempotency. So I think it's fair to call that out here.

guides/dag-best-practices.md
dylanbstorey: functinoally whats the difference between “input” and “set of inputs” ?

Would something like,  “A program is considered idempotent if ,when executed for a given input, produces the same effects if run once or multiple times.” be more succinct ?

kentdanas: The intent is just to convey that this only applies if the input(s) remain the same. So the language should be "set input", rather than "set of inputs". In my mind "set input" and "given input" aren't any different, but I'm willing to be overruled if it's confusing.

collinmcnulty (approved):

guides/dag-best-practices.md
dylanbstorey: makes sense the “set of inputs” was the part throwing me there.

dylanbstorey (approved):

PR #360, opened 2022-04-20
Dynamic task mapping guide

kentdanas:
Hey all - I've got most of a draft for our dynamic task mapping guide here. I still have work to do on the last ML section to get an actual working example, but want to get the ball rolling on reviews since we're hoping to publish next week. Any changes I make to the last section should be mostly limited to the DAG code without much impact on the copy.

@jwitz whenever you have some time, a content review and copy edit as per usual would be much appreciated!

@fhoda @santonaastronomer I'm hoping you could take a look at the last section on using task mapping for ML ops and make sure it all makes sense and tells a good story. Any comments welcome. Mostly looking for a review of the copy, since the DAG isn't done yet, but you can also probably get a sense of where I'm going with the DAG updates.


kentdanas:
@jwitz sorry for the immediate post-vacation ping, but trying to get this out the door within the next couple of days. Even a brief review would be helpful! Thanks in advance!

jwitz:
@kentdanas All good! No lie, the best way to ease back into work is to review someone else's PR 😄  doing that now!

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+## Overview
+
+With the release of Airflow 2.3 (LINK?), users can write DAGs that dynamically generate parallel tasks at runtime. This feature, known as dynamic task mapping, is a paradigm shift for DAG design in Airflow. Prior to Airflow 2.3, tasks could only be generated dynamically at the time the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. Now, you can easily design your DAGs to respond to external criteria by creating a varied number of tasks each time they are run.
+
+In this guide, we'll cover the concept of dynamic task mapping, and a couple of example implementations showing how to use this feature for common use cases.
+
+## Dynamic Task Concepts
+
+Airflow's dynamic task mapping feature it built off of the [MapReduce](https://en.wikipedia.org/wiki/MapReduce) programming model. The map procedure takes a set of inputs and creates a single task for each one. The reduce procedure, which is optional, allows a task to operate on the collected output of a mapped task.
```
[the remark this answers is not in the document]
jwitz: We describe two procedures here and then show two functions in Airflow, but I'm not seeing a clear connection between the descriptions here and the functions. Do these procedures map 1:1 with `expand` and `partial` respectively?

If yes, we should make that more clear in the descriptions of the functions below. If no, I would consider re/moving this paragraph because it is a low-level implementation detail that's difficult to contextualize before learning about the Airflow feature.

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+In practice, DAG authors have two new functions available to implement this feature:
+
+- `expand()`, which passes the input you want to map on to a particular parameter of the operator
+- `partial()`, which passes parameters that stay constant for all mapped instances of the operator
+
+Put together, you can have a task that looks like this:
```
jwitz, line 25:
```suggestion
Tasks have two new functions available to implement this feature:

- `expand()`: This function passes a group of inputs that maps to a single parameter for a given operator. It then creates mapped task runs for each input value. 
- `partial()`: This function passes a group of inputs that apply to all mapped task runs which are generated by `expand()`.

The following task uses both of these functions to dynamically generate 3 task runs:
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+```python
+@task
+    def add(x: int, y: int):
+        return x + y
+
+    added_values = add.partial(y=10).expand(x=[1, 2, 3])
+```
+
+This will result in three mapped `add` tasks, one for each entry in the `x` input list. The `y` parameter remains constant in each task.
```
jwitz, line 35:
```suggestion
This `expand` function creates three mapped `add` tasks, one for each entry in the `x` input list. The `partial` function specifies a value for `y` that remains constant in each task.
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+There are a couple of things to keep in mind when working with mapped tasks:
+
+- You *can* use the results of an upstream task as the input to a mapped task (in fact, this is where the real flexibility comes with this feature). The only requirement is that the upstream task return a value in a dict or list form. If using traditional operators (not decorated tasks), the mapping values must be stored in XCom.
+- You *can* map over multiple parameters. This will result in a cross product with one task for each combination of parameters.
+- You *can* use the results of a mapped task as input to a downstream mapped task.
+- You *can* have a mapped task that results in no task instances (e.g. if your upstream task that generates the mapping values returns an empty list). In this case, the mapped task will be marked skipped, and downstream tasks will be run according to the trigger rules you set (by default, downstream tasks will also be skipped).
+- You *can't* map over all parameters. For example, `task_id`, `pool`, and many `BaseOperator` arguments are not mappable.
```
jwitz, line 43:
```suggestion
- You *can* use the results of an upstream task as the input to a mapped task (in fact, this is one of the most powerful possibilities of this feature). The upstream task must return a value in a `dict` or `list` form. If you're using traditional operators (ie. not [decorated tasks](https://airflow.apache.org/docs/apache-airflow/2.0.0/concepts.html#python-task-decorator)), the mapping values must be stored in XCom.
- You *can* map over multiple parameters. This will result in a cross product with one task for each combination of parameters.
- You *can* use the results of a mapped task as input to a downstream mapped task.
- You *can* have a mapped task that results in no task instances (e.g. if your upstream task that generates the mapping values returns an empty list). In this case, the mapped task will be marked skipped, and downstream tasks will be run according to the trigger rules you set (by default, downstream tasks will also be skipped).
- You *can't* map over specific parameter types. For example, `task_id`, `pool`, and many `BaseOperator` arguments are not mappable.
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+For more high level examples of how to apply dynamic task mapping functions in different cases, check out the Airflow docs (LINK).
```
jwitz, line 45:
```suggestion
For more high level examples of how to apply dynamic task mapping functions in different cases, check out the [Apache Airflow documentation](https://airflow.apache.org/docs/apache-airflow/2.3.0/concepts/dynamic-task-mapping.html).
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+The Airflow UI gives us observability for mapped tasks in both the Graph View and the Grid View.
+
+In the Graph View, any mapped tasks will be indicated by a set of brackets `[ ]` following the task ID. The number in the brackets will update for each DAG run to reflect how many mapped instances were created.
```
jwitz, line 49:
```suggestion
In the **Graph View**, any mapped tasks will be indicated by a set of brackets `[ ]` following the task ID. The number in the brackets will update for each DAG run to reflect how many mapped instances were created.
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+- You *can* map over multiple parameters. This will result in a cross product with one task for each combination of parameters.
+- You *can* use the results of a mapped task as input to a downstream mapped task.
+- You *can* have a mapped task that results in no task instances (e.g. if your upstream task that generates the mapping values returns an empty list). In this case, the mapped task will be marked skipped, and downstream tasks will be run according to the trigger rules you set (by default, downstream tasks will also be skipped).
+- You *can't* map over all parameters. For example, `task_id`, `pool`, and many `BaseOperator` arguments are not mappable.
+
+For more high level examples of how to apply dynamic task mapping functions in different cases, check out the Airflow docs (LINK).
+
+The Airflow UI gives us observability for mapped tasks in both the Graph View and the Grid View.
```
jwitz, line 47:
```suggestion
The Airflow UI gives us observability for mapped tasks in both the **Graph View** and the **Grid View**.
```

jwitz: Let's bold any specific UI elements

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+In the Graph View, any mapped tasks will be indicated by a set of brackets `[ ]` following the task ID. The number in the brackets will update for each DAG run to reflect how many mapped instances were created.
+
+![Mapped Graph](https://assets2.astronomer.io/main/guides/dynamic-tasks/mapped_task_graph.png)
+
+Clicking on the mapped task, we have a new Mapped Instances drop down where we can choose a specific instance to perform task actions on.
```
jwitz, line 53:
```suggestion
Clicking on the mapped task, we have a new **Mapped Instances** drop down where we can select a specific mapped task run to perform actions on. 
```

jwitz: Is the **task actions** page just for the main task, or for each mapped task?

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+![Mapped Actions](https://assets2.astronomer.io/main/guides/dynamic-tasks/mapped_instances_task_actions.png)
+
+Selecting one of the mapped instances provides links to other views like you would see for any other Airflow task: Instance Details, Rendered, Log, XCom, etc.
```
jwitz, line 57: How do you get this selection to appear instead of Instance Details/ All Instances?

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+![Mapped Views](https://assets2.astronomer.io/main/guides/dynamic-tasks/mapped_instance_views.png)
+
+Similarly, the Grid View shows task details and history for each mapped task. All mapped tasks will be combined into one row on the grid (shown as `load_files_to_snowflake [ ]` in this example), and clicking into that task will provide details on each individual mapped instance.
```
jwitz, line 61:
```suggestion
Similarly, the Grid View shows task details and history for each mapped task. All mapped tasks will be combined into one row on the grid (shown as `load_files_to_snowflake [ ]` in the following example). Clicking into that task will provide details on each individual mapped instance.
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+![Mapped Grid](https://assets2.astronomer.io/main/guides/dynamic-tasks/mapped_grid_view.png)
+
+## Example Implementations
+
+In this section we'll show how dynamic task mapping can be implemented for two classic use cases: ELT and ML Ops. The first will highlight use of traditional Airflow operators, and the second will use decorated functions and the TaskFlow API.
```
jwitz, line 67:
```suggestion
In this section we'll show how dynamic task mapping can be implemented for two classic use cases: ELT and ML Ops. The first implementation will use traditional Airflow operators, and the second will use decorated functions and the TaskFlow API.
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+The DAG below completes the following steps:
+
+1. Uses a decorated Python operator to get the current list of files from S3. The S3 prefix passed to this function is parameterized with `ds_nodash` so it pulls files only for the execution date of the DAG run (e.g. for a DAG run on April 12th, we would assume the files landed in a folder named `20220412/`).
+2. Using the results of the first task, map an `S3ToSnowflakeOperator` for each file.
+3. Move the daily folder of processed files into a `processed/` folder while,
+4. Simultaneously (with Step 3), run a Snowflake query that transforms the data. The query is located in a separate SQL file in our `include/` directory. 
+5. Delete the folder of daily files now that it has been moved to `processed/` for record keeping.
```
jwitz, line 79:
```suggestion
1. Use a decorated Python operator to get the current list of files from S3. The S3 prefix passed to this function is parameterized with `ds_nodash` so it pulls files only for the execution date of the DAG run (e.g. for a DAG run on April 12th, we would assume the files landed in a folder named `20220412/`).
2. Using the results of the first task, map an `S3ToSnowflakeOperator` for each file.
3. Move the daily folder of processed files into a `processed/` folder while,
4. Simultaneously (with Step 3), run a Snowflake query that transforms the data. The query is located in a separate SQL file in our `include/` directory. 
5. Delete the folder of daily files now that it has been moved to `processed/` for record keeping.
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+
+    copy_to_snowflake >> [move_s3, transform_in_snowflake]
+    move_s3 >> delete_landing_files
+```
+
+The Graph View of the DAG looks like this:
+
+![ELT Graph](https://assets2.astronomer.io/main/guides/dynamic-tasks/mapping_elt_graph.png)
+
+Keep in mind the format needed for the parameter you are mapping on. In the example above, we write our own Python function to get the S3 keys because the `S3toSnowflakeOperator` requires *each* `s3_key` parameter to be in a list format, and the `s3_hook.list_keys` function returns a single list with all keys. By writing our own simple function, we can turn the hook results into a list of lists that can be used by the downstream operator.
```
jwitz, line 143:
```suggestion
When dynamically mapping tasks, make note of the format needed for the parameter you are mapping on. In the example above, we write our own Python function to get the S3 keys because the `S3toSnowflakeOperator` requires *each* `s3_key` parameter to be in a list format, and the `s3_hook.list_keys` function returns a single list with all keys. By writing our own simple function, we can turn the hook results into a list of lists that can be used by the downstream operator. 
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,363 @@
...
+            'solver': ['newton-cg', 'lbfgs', 'liblinear', 'sag', 'saga']
+        }
+    }  
+}
+
+```
+
+The model information could come from any external system as well, with an update to the `get_models()` task. The only requirement is that the resulting map input is a dict or list.
+
+Also note that for decorated tasks like in the DAG above, the mapping parameters are automatically passed by calling the proper functions, leveraging the TaskFlow API to avoid explicit calling of XCom.
```
jwitz, line 363: Maybe add a line here or before the code block about how this differs from the first example? From what I can tell, it means that we don't need to implement the custom functions that change the value format in the first example, right? Could be nice to include an equivalent note in the first example ("To avoid needing to typecast your parameters like this, do blah blah blah like in the following example")

jwitz (changes requested):
Overall looks good @kentdanas ! Requesting changes mainly on our first general explainer of task mapping - I'd like to make it more clear.

Is it worth explaining the difference between this feature and[ dynamic DAGs](https://www.astronomer.io/guides/dynamically-generating-dags/#:~:text=Another%20way%20to%20define%20input,that%20source%20avoids%20redundant%20work.)? The use similar language, but seem like very different implementations.

guides/dynamic-tasks.md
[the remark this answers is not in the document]
kentdanas: No - `expand` does map (for lack of a better word) to the map procedure in MapReduce. Reduce just means you can take the output of all your mapped tasks and "reduce" back down to a single task afterwards - so in Airflow, you can have a single downstream task that is dependent on the results of your mapped task (but you don't have to).

`partial` is just a way to pass consistent parameters in to every task within a set of mapped tasks.

I don't want to remove the paragraph since this was a really important distinction is terms of _how_ dynamic tasks were implemented in the context of Airflow, but I'll try to make it clearer.

guides/dynamic-tasks.md
kentdanas: Both - you can perform task actions on the main task (every mapped instance), or on individual mapped tasks

guides/dynamic-tasks.md
kentdanas: It appears automatically for any mapped tasks

guides/dynamic-tasks.md
kentdanas: The only reason we had to change the mapping value format in the first example was because of the operator we used (S3toSnowflake). That isn't relevant at all in this example since we're using different operators here that don't require that. But that's only related to the operators chosen, not the high level use case or example, so it wouldn't make sense to call it out.

The only real difference between the two examples is the use case.

guides/dynamic-tasks.md
kentdanas: I changed your edits a bit since they weren't quite right (although I understand how you got there based on my original copy). Let me know if the new definitions are clearer.

kentdanas:
Thanks @jwitz! I made some changes to the concepts section, let me know if you think it's clearer now. 

As far as dynamic DAGs - they are related conceptually (both just mean some part of your DAG is changing based on external factors), but not at all related in terms of the actual features or implementation. In general, when to use one or the other will be more use case driven, and I think might be better suited towards covering in other content (I'll talk about it in our webinar, and we will have a blog post that talks about benefits of dynamic tasks).

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,367 @@
...
+                    early_stopping_rounds=5
+                )
+
+            else:
+                logging.info('Training model with best parameters')
+                clf = LogisticRegression(penalty=best_params['penalty'], C=best_params['C'], solver=best_params['solver']).fit(X_train, y_train)
+
+            y_pred_class = metrics.test(clf, X_test)
+
+            # Log Classfication Report, Confusion Matrix, and ROC Curve
```
jwitz, line 324:
```suggestion
            # Log Classification Report, Confusion Matrix, and ROC Curve
```

guides/dynamic-tasks.md
```diff
@@ -0,0 +1,367 @@
...
+
+Prior to Airflow 2.3, tasks could only be generated dynamically at the time that the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. With dynamic task mapping, you can easily write DAGs that create tasks based on your current runtime environment. 
+
+In this guide, we'll explain the concept of dynamic task mapping and provide example implementations for common use cases.
+
+## Dynamic Task Concepts
+
+Airflow's dynamic task mapping feature is built off of the [MapReduce](https://en.wikipedia.org/wiki/MapReduce) programming model. The map procedure takes a set of inputs and creates a single task for each one. The reduce procedure, which is optional, allows a task to operate on the collected output of a mapped task. In practice, this means that your DAG can create an arbitrary number of parallel tasks at runtime based on some input parameter(s) (the "map"), and then if needed, have a single task downstream of your parallel mapped tasks that depends on their output (the "reduce").
```
jwitz, line 20: EXCELLENT rewrite here!

jwitz (approved):
Just one minor spelling thing, otherwise LGTM! Agreed that it doesn't make a lot of sense to compare and contrast with dynamic DAGs - I just brought it up in case you anticipate users being confused given the overloading of "dynamic" in two different Airflow context. If you think that's likely, it might be worth adding a single sentence being like "hey, this ISN'T related to Dynamic DAGs, those are two separate implementations of dynamic generation for different use cases."

PR #374, opened 2022-05-17
Update airflow-openlineage.md with more complete example

RNHTTR:
I'm not sure if it's something with my version of `astrocloud`, but I don't get a connection called `postgres_default` when I run `astrocloud dev init`, so I think it makes sense to include instructions for creating a connection.

I also converted the SQL for creating & populating the source tables from raw SQL into a DAG. If it makes more sense to leave it as SQL, I think that's fine as long as we include how to run those queries against the metadata db.


RNHTTR:
Closing as addressed by @kentdanas far more elegant changes in #375

PR #376, opened 2022-05-20
First draft updated guide on hooks

TJaniF:
This is the first draft of an updated version of the Hooks 101 guide including instructions how to install hooks, a description of the S3Hook including its currently available methods, and an example DAG using the S3Hook and the SlackHook.

### DIFF
--- a/guides/what-is-a-hook.md
+++ b/guides/what-is-a-hook.md
@@ -144,6 +144,6 @@ with DAG(dag_id='hook_tutorial',
 
 The DAG above completes the following steps:
 
-1. Use a decorated Python operator with a manually implemented S3Hook to read three specific keys from S3 and return a dictionary with the file contents converted to integers.
+1. Use a decorated Python operator with a manually implemented S3Hook to read three specific keys from S3 with the `read_key` method. Return a dictionary with the file contents converted to integers.
 2. Using the results of the first task a second decorated Python operator performs a simple sum check. 
-3. Lastly, a SlackHook is used within a third task to post the result of the check to a Slack Channel and return the response from the Slack API.
+3. Post the result of the check to a Slack Channel using the `call` method of the SlackHook and return the response from the Slack API.
--- a/guides/what-is-a-hook.md
+++ b/guides/what-is-a-hook.md
@@ -131,7 +131,7 @@ def post_to_slack(sum_check_result):
 						json={"channel": "#test-airflow",
 						"text": f"""A test on your bucket contents failed!
 						Target sum not reached: {sum_check_result[1]}"""})
-
+						
 		# return the response of the API call to log it or use it downstream
     return server_response
 
--- a/guides/what-is-a-hook.md
+++ b/guides/what-is-a-hook.md
@@ -23,7 +23,7 @@ Hooks wrap around APIs and provide methods to interact with different external s
 To use a hook, you typically only need a connection ID to connect with an external system. More information on how to set up connections can be found in [Managing your Connections in Apache Airflow](https://www.astronomer.io/guides/connections/) or in the example section below.
 
 All hooks inherit from the [BaseHook class](https://github.com/apache/airflow/blob/main/airflow/hooks/base.py), which contains the logic to set up an external connection given a connection ID.
-On top of making the connection to an external system, each Hook might contain additional methods to perform various actions within that system. These methods might rely on different Python libraries for these interactions.
+On top of making the connection to an external system, each hook might contain additional methods to perform various actions within that system. These methods might rely on different Python libraries for these interactions.
 
 For example, the [`S3Hook`](https://registry.astronomer.io/providers/amazon/modules/s3hook), which is one of the most widely used hooks, relies on the [`boto3`](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) library to manage its connection with S3.  
 
--- a/guides/what-is-a-hook.md
+++ b/guides/what-is-a-hook.md
@@ -125,12 +125,12 @@ def post_to_slack(sum_check_result):
         server_response = slack_hook.call(api_method='chat.postMessage',
 						json={"channel": "#test-airflow",
 						"text": f"""All is well in your bucket!
-										Correct sum: {sum_check_result[1]}!"""})
+						Correct sum: {sum_check_result[1]}!"""})
     else:
         server_response = slack_hook.call(api_method='chat.postMessage',
 						json={"channel": "#test-airflow",
 						"text": f"""A test on your bucket contents failed!
-										Target sum not reached: {sum_check_result[1]}"""})
+						Target sum not reached: {sum_check_result[1]}"""})
 
 		# return the response of the API call to log it or use it downstream
     return server_response
--- a/guides/what-is-a-hook.md
+++ b/guides/what-is-a-hook.md
@@ -144,6 +144,6 @@ with DAG(dag_id='hook_tutorial',
 
 The DAG above completes the following steps:
 
-1. Use a decorated Python operator with a manually implemented S3Hook to read three specific keys from S3 with the `read_key` method. Return a dictionary with the file contents converted to integers.
-2. Using the results of the first task a second decorated Python operator performs a simple sum check. 
+1. Uses a decorated Python Operator with a manually implemented `S3Hook` to read three specific keys from S3 with the `read_key` method. Returns a dictionary with the file contents converted to integers.
+2. Using the results of the first task, a second decorated Python Operator performs a simple sum check. 
 3. Post the result of the check to a Slack Channel using the `call` method of the SlackHook and return the response from the Slack API.
--- a/guides/what-is-a-hook.md
+++ b/guides/what-is-a-hook.md
@@ -1,194 +1,57 @@
 ---
 title: "Hooks 101"
 description: "An introduction to Hooks in Apache Airflow."
-date: 2018-05-21T00:00:00.000Z
+date: 2022-05-24T00:00:00.000Z
 slug: "what-is-a-hook"
 heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_preview.png"
 tags: ["Hooks", "Operators", "Tasks", "Basics"]
 ---
 
+## Overview
 
-## Hooks
+Hooks are one of the fundamental concepts of Airflow. At a high level, a hook is an abstraction of a specific API that makes it possible for Airflow to interact with an external system. Hooks are built into many operators, but they can also be used directly in DAG code.
 
-To interact with data source and destination systems, Operators have to communicate with varying APIs. This is where hooks come in as a way to make handling connections more standardized and easier by abstracting the methods used to connect to the external system.
-
-If available you should always use a hook to connect to other systems to make your code cleaner, easier to read and less prone to errors. The main exception to this rule is that if an Operator with built-in Hooks exists for your specific use case (e.g [S3ToSnowflakeOperator](https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/operators/s3_to_snowflake.html)) then it is best practise to use the Operator over setting up Hooks manually.
-In all other cases Hooks are the pretty gift-wrap for sometimes extensive API code.
+In this guide we'll cover the basics of using hooks in Airflow, when to use them directly in DAG code and provide you with an example DAG using two different hooks.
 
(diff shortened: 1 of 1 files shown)


guides/what-is-a-hook.md
```diff
@@ -7,91 +7,257 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
 tags: ["Hooks", "Operators", "Tasks", "Basics"]
 ---
 
+
 ## Hooks
```
kentdanas, line 11: Love the content in this! I think we can rearrange it a bit to be consistent with some of our other guides in terms of flow/sections (not all of them follow a pattern, but we try to be consistent with the newer ones). Check out the Sensors 101 guide as an example: https://www.astronomer.io/guides/what-is-a-sensor

I'm thinking for this one we could have:

- Overview (basically your intro, plus a short "in this guide, we'll cover....")
- Hook Basics (basically your Example: S3Hook section, but maybe a bit more general. More comments on this below)
- When to Use Hooks (short section describing practically when hooks are used e.g. if an operator doesn't exist, etc. these can be bullet points without code examples)
- Example Implementation (the example DAG you provide)

guides/what-is-a-hook.md
```diff
@@ -7,91 +7,257 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
 
-Operators are the main building blocks of Airflow, but operators rely heavily upon Hooks to interact with all of their source and destination systems.
+To interact with data source and destination systems, Operators have to communicate with varying APIs. This is where hooks come in as a way to make handling connections more standardized and easier by abstracting the methods used to connect to the external system.
```
kentdanas, line 13: Maybe we can rewrite this first paragraph to be a bit more "hook-forward", something like "[Hooks](https://airflow.apache.org/docs/apache-airflow/stable/concepts/connections.html#hooks) are one of the most basic concepts of Airflow. At a high level, hooks are an abstraction of an API that allows Airflow to interface with an external system. Hooks are the building blocks of operators, but they can also be used themselves in DAG code." 

Or something like that, and then have a second short paragraph that describes what we'll cover in the guide.

guides/what-is-a-hook.md
```diff
@@ -7,91 +7,257 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
+
+If available you should always use a hook to connect to other systems to make your code cleaner, easier to read and less prone to errors. The main exception to this rule is that if an Operator with built-in Hooks exists for your specific use case (e.g [S3ToSnowflakeOperator](https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/operators/s3_to_snowflake.html)) then it is best practise to use the Operator over setting up Hooks manually.
+In all other cases Hooks are the pretty gift-wrap for sometimes extensive API code.
+
+>[Over 200 Hooks](https://registry.astronomer.io/modules/?types=hooks%2CHooks&page=2) are currently listed in the Astronomer Registry.  If there isn't one for your use-case yet you can of course write your own and are welcome to share it with the community!
+
+<br>
 
-Hooks are used as a way to abstract the methods you would use against a source system, but often contain limited logic for how that system is interacted with. Hooks should always be used to connect with any external system.
+### Install
```
kentdanas, line 22: I think we can move this down within the Example Implementation section where you have the example DAG below. It's definitely worth noting that you have to install relevant provider packages to use most hooks, but probably doesn't need it's own high level section in the guide.

guides/what-is-a-hook.md
```diff
@@ -7,91 +7,257 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
 
 ```python
+from airflow.providers.<provider>.<subpackages> import <Hook>
+```
+
+You will also need to set up a connection with your external system. An example for an AWS S3 bucket and for Slack is described below.
+
+<br>
+
+### Example: S3Hook
```
kentdanas, line 42: Let's rewrite this section to be more general "Hook Basics". I think the most important things to cover are:

- Hooks are abstractions on APIs
- Hooks are used to make connections to external systems
- Hooks can also have other methods for interacting with those systems, and these will vary by each hook

A lot of this is already in here, but I think we can make it less S3 specific and a little clearer. Also, I don't think we need to show every method in the Hook (I know it was this way before, but I don't think it's super helpful). Maybe just highlight a few of the additional methods as an example.

guides/what-is-a-hook.md
```diff
@@ -7,91 +7,257 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
+
+	def delete_bucket_tagging(self, bucket_name=None):
+		return None
+
+```
+
+<br>
+
+### Simple example:  DAG with S3Hook and SlackHook
+
```
kentdanas, line 183: This is a great example! Let's describe up front why you would want to write a DAG this way using Hooks (as opposed to using specific operators) - I'm guessing because you can't pass information between S3ListKeys (or whatever it's called) operator and SlackWebhookOperator in the way you're showing here, but we should make that explicit

kentdanas (changes requested):
@TJaniF this is a great start! I left some comments throughout, mostly about high level organization to pull this more in line with some of our other (recent) guides. Let me know if you have questions about any of the requested updates.

guides/what-is-a-hook.md
TJaniF: I expanded the example to include a third task in between the two Hooks that uses the output from the S3Hook (that now reads three files from the S3) and provides input to the SlackHook, as well as added in the return of the Slack API response because otherwise one could just use the SlackAPIPostOperator.

TJaniF:
I rearranged the content in the way suggested and expanded the example DAG so the use-case is not covered by existing Operators.

guides/what-is-a-hook.md
```diff
@@ -1,97 +1,149 @@
...
 
-    def delete_objects(self, bucket, keys):
-        return response
+```text
+apache-airflow-providers-amazon
+apache-airflow-providers-slack
 ```
 
- This Hook inherits from the [AwsBaseHook](https://registry.astronomer.io/providers/amazon/modules/awsbasehook) which inherits from the [BaseHook](https://registry.astronomer.io/providers/apache-airflow/modules/basehook). All Hooks inherit from the `BaseHook` which contains the logic for how hooks interact with `Airflow Connections`. `Connections`  are Airflow's built in credential-store for your source/destination systems. Hooks are designed to handle these in a clean, reusable way. Tasks that use a hook will have an input parameter for the `conn_id` of the connection you wish to use.
+Next you will need to set up the connection to the S3 bucket and Slack in the Airflow UI (localhost:8080 by default).
```
kentdanas, line 70:
```suggestion
Next you will need to set up the connection to the S3 bucket and Slack in the Airflow UI.
```

guides/what-is-a-hook.md
```diff
@@ -1,97 +1,149 @@
...
+        start_date=datetime(2022,5,20),
+        schedule_interval='@daily',
+        catchup=False,
+        ) as dag:
+
+    # the dependencies are automatically set by XCom
+    response = read_keys_form_s3()
+    sum_check_result = run_sum_check(response)
+    post_to_slack(sum_check_result)
+```
```
kentdanas, line 149: Maybe just add one final sentence or two here describing the specific Hooks used for accomplishing the use case? Check out this guide as an example (it describes the DAG's steps before the code, which would also be fine): https://www.astronomer.io/guides/dynamic-tasks#example-implementations

kentdanas (changes requested):
@TJaniF this is looking really good! Thanks for your updates!

Overall I think the content here is good. I went through and made some suggested edits, mostly small things. (I suggest reviewing them from the `Files changed` tab and batching them, so you don't have a commit and trigger the CI for every single one :) )

I'll also request a copy edit from @jwitz who usually reviews our guides before we publish them.

guides/what-is-a-hook.md
kentdanas: Looks great!

guides/what-is-a-hook.md
```diff
@@ -1,97 +1,149 @@
 ---
 title: "Hooks 101"
 description: "An introduction to Hooks in Apache Airflow."
-date: 2018-05-21T00:00:00.000Z
+date: 2022-05-24T00:00:00.000Z
```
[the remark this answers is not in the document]
TJaniF: That does make sense, I changed it back :)

guides/what-is-a-hook.md
TJaniF: I added a description of the DAGs steps at the end of the document and included which Hooks were used with which methods there.

kentdanas (approved):
Thanks @TJaniF, this looks great! One more minor comment, otherwise I'll hand off to Jake for a copy edit.

@jwitz you mind giving this a read when you have a minute?

guides/what-is-a-hook.md
```diff
@@ -7,91 +7,143 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
+3. For the connection to Slack select 'Slack Webhook' as the connection type and provide your [Bot User OAuth Token](https://api.slack.com/authentication/oauth-v2) as a password. This token can be obtained by navigating to the 'OAuth & Permissions tab' under 'Features' on api.slack.com/apps.
 
- **Hooks provide an interface in which to interact with an external system, but do not contain the logic for how that system is interacted with.** (in this example, boto3 holds the logic on how to interact with S3)
+The DAG below uses [Airflow Decorators](https://registry.astronomer.io/guides/airflow-decorators) to define tasks and [XCom](https://registry.astronomer.io/guides/airflow-passing-data-between-tasks) to pass information between them. The name of the S3 bucket and the names of the files that the first task reads were stored as environment variables for security purposes.
```
jwitz, line 71:
```suggestion
The DAG below uses [Airflow Decorators](https://registry.astronomer.io/guides/airflow-decorators) to define tasks and [XCom](https://registry.astronomer.io/guides/airflow-passing-data-between-tasks) to pass information between them. The name of the S3 bucket and the names of the files that the first task reads are stored as environment variables for security purposes.
```

jwitz, line 71:
```suggestion
The DAG below uses [Airflow Decorators](https://registry.astronomer.io/guides/airflow-decorators) to define tasks and [XCom](https://registry.astronomer.io/guides/airflow-passing-data-between-tasks) to pass information between them. The name of the S3 bucket and the names of the files that the first task reads are stored as environment variables for security purposes.
```

guides/what-is-a-hook.md
```diff
@@ -7,91 +7,143 @@ heroImagePath: "https://assets.astronomer.io/website/img/guides/IntroToDAG_previ
...
+```
 
-- `https://github.com/apache/airflow/tree/master/airflow/hooks`
-- `https://github.com/apache/airflow/tree/master/airflow/contrib/hooks`
+The DAG above completes the following steps:
 
-Of course, you can write hooks to interact with any system you want!
+1. Uses a decorated Python Operator with a manually implemented `S3Hook` to read three specific keys from S3 with the `read_key` method. Returns a dictionary with the file contents converted to integers.
+2. Using the results of the first task, a second decorated Python Operator performs a simple sum check. 
+3. Post the result of the check to a Slack Channel using the `call` method of the SlackHook and return the response from the Slack API.
```
jwitz, line 149:
```suggestion
1. Use a decorated Python Operator with a manually implemented `S3Hook` to read three specific keys from S3 with the `read_key` method. Returns a dictionary with the file contents converted to integers.
2. With the results of the first task, use a second decorated Python Operator to complete a simple sum check. 
3. Post the result of the check to a Slack channel using the `call` method of the SlackHook and return the response from the Slack API.
```

jwitz, line 149:
```suggestion
1. Use a decorated Python Operator with a manually implemented `S3Hook` to read three specific keys from S3 with the `read_key` method. Returns a dictionary with the file contents converted to integers.
2. With the results of the first task, use a second decorated Python Operator to complete a simple sum check. 
3. Post the result of the check to a Slack channel using the `call` method of the SlackHook and return the response from the Slack API.
```

jwitz (approved):
@TJaniF This guide is great! Such a huge improvement over our existing Hooks guide - I was able to immediately figure out why Hooks are important and how to use them.

Don't be alarmed by the number of suggestions here, basically all of them are minor copy changes. Feel free to tag me again if you have any questions or if one of my suggestions missed the mark.

TJaniF:
@jwitz Thank you! :)

PR #377, opened 2022-05-26
Remove images that are 404-ing

allenbrady-astro:
This PR removes broken images from the [Editing Task and DAG Metadata](https://www.astronomer.io/guides/editing-task-dag-metadata) guide to fix SEO issue. Couldn't find the originals but looks like they were being pulled in from `assets.astronomer.io` which is now defunct.


kentdanas (approved):
Thanks @allenbrady-astro! It's possible we will take this guide down in the future (pretty sure it isn't recommending best practices). But for now all good to just remove the images.

allenbrady-astro:
Sounds good, thank you @kentdanas!

PR #381, opened 2022-06-02
Add Astro link to error-notifications-in-airflow

magdagultekin:
In the section `Notifications on Astronomer` we don't provide a link for Astro, only for Software. I thought it would be a good idea to add it, what do you think? 

Additionally, the sentence re providers in Airflow 2.0+ is repeated twice - removed one of them and moved the note to the proper point.


kentdanas (approved):
These updates look great, thanks @magdagultekin!

PR #383, opened 2022-06-03
Update airflow-dbt.md

magdagultekin:
Fixed a typo and added `dag_id=` for better readability.


kentdanas:
Thanks @magdagultekin! re: your question - no, it’s not considered top-level as a Jinja expression. The DAG parser considers that a regular string. Jinja expressions/templates aren’t evaluated until right before the execute() method of the task is called. So the value passed to the operator is still a string of "{{ var.value.dbt_cloud_job_id }}" until the template gets rendered during task execution.

If the variable assignment was JOB_ID = Variable.get("dbt_cloud_job_id") then the parser itself would execute that code, connect the metadatabase and fetch the Variable record. Then it’s considered top-level code.

kentdanas (approved):

magdagultekin:
Awesome, thank you very much @kentdanas!

PR #385, opened 2022-06-06
Database updated guide first draft

TJaniF:
First draft of a shortened and updated guide to the metadata database in Airflow. The sections are: short intro and disclaimer, content description (without schemas), best practices and disaster recovery.


guides/airflow-database.md
```diff
@@ -8,640 +8,53 @@ tags: ["Database", "SQL", "Components"]
...
+## Airflow Metadata Database Best Practices
 
--   [public.ab\_permission\_view](#public.ab\_permission\_view)
+1. When up- or downgrading Airflow, which commonly includes changes to your metadata database make sure you always have a backup of your metadata database and follow the recommended steps for changing Airflow versions: backup the database, check for deprecated features, pause all DAGs and make sure no tasks are running.
 
-------------------------------------------------------------------------
+2. Use caution when [pruning old records](https://airflow.apache.org/docs/apache-airflow/stable/usage-cli.html#purge-history-from-metadata-database) from your database with `db clean`.
 
-### public.ab\_permission\_view
+3. Keep in mind that every time you access data from the metadata database from within a DAG (for example by fetching a variable, pulling from XCom or using a connection ID) you will make a connection to the metadata database which needs compute resources. It is therefore best practise to try to keep these actions within tasks so these connections are not created every time the scheduler parses the DAG file (every 30 seconds!).
```
kentdanas, line 48:
```suggestion
3. Keep in mind that every time you access data from the metadata database from within a DAG (for example by fetching a variable, pulling from XCom or using a connection ID) you will make a connection to the metadata database which needs compute resources. It is therefore best practice to try to keep these actions within tasks, which will only create a connection to the database at run time. If they are written as top level code, connections will be created every time the scheduler parses the DAG file (every 30 seconds by default!).
```

guides/airflow-database.md
```diff
@@ -8,640 +8,53 @@ tags: ["Database", "SQL", "Components"]
...
 
-**Table Structure:**
+4. Memory in the Airflow metadata database is limited depending on your setup. Astro CLI uses a Postgres database of the size of 1GB while local Airflow by default will use a 2GB SQLite database. This is one of the many reasons why we highly advise against moving large amounts of data via XCom.
```
kentdanas, line 50:
```suggestion
4. Memory in the Airflow metadata database can be limited depending on your setup. For example, local options will generally have lower memory: Astro CLI uses a Postgres database of the size of 1GB while local Airflow by default will use a 2GB SQLite database. Running low on memory in your metadata database can cause performance issues in Airflow. This is one of the many reasons why we highly advise against moving large amounts of data via XCom, and recommend using a cleanup and archiving mechanism in any production deployments.
```

Let me know if this makes sense. Wanted to make the distinction between memory considerations and using a managed db service (related, but separate best practices).

guides/airflow-database.md
```diff
@@ -8,640 +8,53 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+5. Because of limited memory in the metadata database it is recommended to use a cleanup and archive mechanism in deployment (for example [Amazon Redshift](https://aws.amazon.com/redshift/)).
```
kentdanas, line 52:
```suggestion
5. Since the metadata database is critical for the scalability and resiliency of your Airflow deployment, it is best practice at least for production Airflow deployments to use a managed database service, for example [AWS RDS](https://aws.amazon.com/rds/) or [Google Cloud SQL](https://cloud.google.com/sql).
```

kentdanas (changes requested):
@TJaniF great start with this! Left a few comments throughout, and as we discussed this morning, I think we can add a section on "interacting with the metadata database" that covers the SQLAlchemy recommendations and using the UI/API so we can get rid of that queries guide. Let me know if you have questions on anything and we can go from there :)

guides/airflow-database.md
```diff
@@ -8,640 +8,53 @@ tags: ["Database", "SQL", "Components"]
...
+In this guide we will explain what kind of data Airflow saves in its metadata database, give an overview over key tables and Airflow specific database best practises, as well as talk about disaster recovery in Airflow.
 
-Even though we don't recommend modifying the values directly on the database, as doing so might affect dependencies, understanding the underlying structure can be very useful when it comes to building your own reports or querying directly against the database. You can find some useful queries that go directly against these tables in our [Useful SQL queries for Apache Airflow](https://www.astronomer.io/guides/airflow-queries/) guide. 
+## Content of the Metadata Database
 
+Airflow uses SQLAlchemy and Object Relational Mapping (ORM) in Python to connect and interact with the underlying metadata database from the application layer. Thus, any database supported by [SQLAlchemy](https://www.sqlalchemy.org/) can theoretically be configured to host Airflow's metadata. On Astronomer, each Airflow deployment is equipped with PostgreSQL database for this purpose.
```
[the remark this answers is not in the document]
TJaniF: Yes, I think that makes a lot of sense (I had that info in point 4 of best practices)!

guides/airflow-database.md
```diff
@@ -8,640 +8,53 @@ tags: ["Database", "SQL", "Components"]
...
 
-## ERD Diagram
+There are several types of metadata stored in the metadata database. One group of tables handles users' login information and permissions for a specific airflow instance, another set of tables deals with storing variables, connections and XComs while other tables contain data the scheduler uses to organize DAG and task runs.
```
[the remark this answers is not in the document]
TJaniF: The tables that handle user and permission data all have the "ab_" prefix. The other two groups "storing input" (variables, connections, XComs) vs scheduler organizing tables were categories I used to break up the information description. I converted the paragraph to bullet points.

guides/airflow-database.md
```diff
@@ -8,640 +8,53 @@ tags: ["Database", "SQL", "Components"]
...
-|----------|-------|--------|------------------|
-|          | id    |integer | *PRIMARY KEY*    |
-|          | name  |character varying(100)  | *UNIQUE NOT NULL* |
+There are many more tables in the metadata database storing data ranging from DAG tags over serialized DAG code to current states of sensors and rescheduled tasks.
 
-**Tables referencing this one via Foreign Key Constraints:**
+## Airflow Metadata Database Best Practices
 
--   [public.ab\_permission\_view](#public.ab\_permission\_view)
+1. When up- or downgrading Airflow, which commonly includes changes to your metadata database make sure you always have a backup of your metadata database and follow the recommended steps for changing Airflow versions: backup the database, check for deprecated features, pause all DAGs and make sure no tasks are running.
```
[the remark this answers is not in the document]
TJaniF: There is a doc page (https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading.html?highlight=upgrade) but it does not mention to check for deprecated features and to pause the DAGs/have no tasks running, those are best practices that Marc talked about in the Fundamentals course. Maybe they are obvious to experienced users? I thought it makes sense to include Marc's steps just in case.

guides/airflow-database.md
```diff
@@ -8,640 +8,127 @@ tags: ["Database", "SQL", "Components"]
...
-|       |id   |integer |   *PRIMARY KEY*|
-|[public.ab\_permission\_view.id](#public.ab\_permission\_view)   |permission\_view\_id   |integer   |*UNIQUE\#1*|
-|[public.ab\_role.id](#public.ab\_role)                         |role\_id              |integer   |*UNIQUE\#1*|
+## Interacting with the Metadata Database
 
-------------------------------------------------------------------------
+> **Note**: For many use-cases you can access contents from the metadata database via the Airflow UI or the stable REST API. These points of access are always preferable to querying the metadata database directly!
 
-### public.ab\_register\_user
+For use cases where neither the Airflow UI nor the REST API can provide sufficient data it is recommended to use SQLAlchemy to query the metadata database.
```
kentdanas, line 74:
```suggestion
The preferred methods for retrieving data from the metadata database are using the Airflow UI or making a request to the Airflow API. The UI and API provide viewing access to much of the data in the metadata database, and accessing data using them reduces the potential of accidentally making changes to data that can harm your Airflow instance. For use cases where neither the Airflow UI nor the REST API can provide sufficient data it is recommended to use SQLAlchemy to query the metadata database.
```

I didn't write this particularly well so you may want to reword, but I think we can say something like this and then remove the note above which is a bit repetitive. Also worth adding a link to the API docs.

kentdanas, line 74: Can we add a sentence about why we recommend SQLAlchemy? (hoping that slack thread has some reasoning in it...)

guides/airflow-database.md
```diff
@@ -8,640 +8,127 @@ tags: ["Database", "SQL", "Components"]
...
+        FROM alembic_version;"""
 
-**Table Structure:**
+with Session(engine) as session:
+    result = session.execute(stmt)
+    print(result.all()[0][0])
+```
 
-------------------------------------------------------------------------
+## Example: Retrieving Number of Successfully Completed Tasks
```
kentdanas, line 94: I wonder if we can make this section more general "Example Queries" since we're going to remove the queries guide, and then have a few bullet points or small section headers that cover those metadata "queries" and state the preferred method for executing (UI/API, or SQLAlchemy). Then for the alembic example, it could serve as the bullet point for which you have to use SQLAlchemy, but not be such a standalone example. Let me know if you want to talk through this tomorrow.

kentdanas (commented):
@TJaniF this is looking really good! A couple of additional comments, most of them minor wording changes. Also have a few things about the examples that I think we can expand on, we can chat more about this tomorrow.

guides/airflow-database.md
```diff
@@ -8,640 +8,127 @@ tags: ["Database", "SQL", "Components"]
...
-### public.ab_permission
- 
-**Table Structure:**
+There are several types of metadata stored in the metadata database.
 
-------------------------------------------------------------------------
+- Tables related to users' login information and permissions for a specific airflow instance. The names of these tables all start with `ab_`.
+- Tables storing input like variables, connections and XComs.
+- Tables the scheduler uses to organize DAG and task runs.  
+- Other tables for example to keep track of the current [Alembic](https://alembic.sqlalchemy.org/en/latest/index.html) version or of tasks that missed their [SLA](https://www.astronomer.io/guides/error-notifications-in-airflow/#airflow-slas).
```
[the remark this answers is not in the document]
TJaniF: The alembic version is actually the only kind of version id (other than session ID) I found stored in the metadatabase, but I reworded the 4th bullet point with more general examples. Yes, SLA definitely falls under point 3, sorry about that.

guides/airflow-database.md
TJaniF: So this sent me on a small treasure hunt today because the way I was using it, it did not actually provide any advantage over direct SQL. 
SQLAlchemy offers the possibility to query from defined models in the Airflow source code, the same way the webserver does, which gives the advantage that changes to the schema should not break the sqlalchemy code unless the models themselves were majorly changed because schema changes get added to the models.

kentdanas: Ahhh that's right, I had heard that a long time ago and couldn't remember exactly what the reasoning was :) Thanks for tracking that down, that will definitely be good info to include!

guides/airflow-database.md
```diff
@@ -8,640 +8,234 @@ tags: ["Database", "SQL", "Components"]
...
-|                                                      |id              |integer   |*PRIMARY KEY*|
-|  [public.ab\_permission.id](#public.ab\_permission)   |permission\_id   |integer   |*UNIQUE\#1*|
-|  [public.ab\_view\_menu.id](#public.ab\_view\_menu)    |view\_menu\_id    |integer   |*UNIQUE\#1*|
+- [Variables](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html)
+- [Connections](https://www.astronomer.io/guides/connections)
+- [XComs](https://www.astronomer.io/guides/airflow-passing-data-between-tasks)
+- [Pools](https://www.astronomer.io/guides/airflow-pools/)
 
-**Tables referencing this one via Foreign Key Constraints:**
+The information in these tables can be viewed and modified under the **Admin** tab in the Airflow UI or by using the Airflow REST API.
```
kentdanas, line 65: Let's link to the REST API docs here. I'll leave it up to you whether it makes sense to link to specific endpoints for the content you mention. If they're easy to find (I suspect these ones are), then one general link should be fine.

guides/airflow-database.md
```diff
@@ -8,640 +8,234 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+4. Memory in the Airflow metadata database can be limited depending on your setup and running low on memory in your metadata database can cause performance issues in Airflow. This is one of the many reasons why we highly advise against moving large amounts of data via XCom, and recommend using a cleanup and archiving mechanism in any production deployments.
 
-### public.ab_role
+5. Since the metadata database is critical for the scalability and resiliency of your Airflow deployment, it is best practice at least for production Airflow deployments to use a managed database service, for example [AWS RDS](https://aws.amazon.com/rds/) or [Google Cloud SQL](https://cloud.google.com/sql).
 
-**Table Structure:**
+## Example Queries
```
kentdanas, line 102:
```suggestion
## Examples Methods for Accessing Data of Interest
```
Bringing this in line with the language in the overview section last bullet. Alternatively, if you'd prefer to keep it as 'Example Queries' you could update it back in both places. I think saying 'queries' is still relevant since that's what the UI/API are doing for you.

guides/airflow-database.md
```diff
@@ -8,640 +8,234 @@ tags: ["Database", "SQL", "Components"]
...
+```
 
-------------------------------------------------------------------------
+### Example: Delete a DAG
 
-|F-Key   |Name   |Type                     |Description
-|------ |----- |----------------------- |-------------------
-|        |id     |integer                  |*PRIMARY KEY*
-|        |name   |character varying(100)   |*UNIQUE NOT NULL*
+Deleting the metadata of a DAG can be accomplished either by clicking the trashcan icon in the Airflow UI or sending a DELETE request via the Airflow REST API. This is not possible while the DAG is still running and will not delete the python file in which the DAG is defined, meaning the DAG will appear again in your UI with no history at the next parsing of the `/dags` folder from the scheduler.
```
kentdanas, line 167:
```suggestion
Deleting the metadata of a DAG can be accomplished either by clicking the `trashcan` icon in the Airflow UI or sending a `DELETE` request via the Airflow REST API. This is not possible while the DAG is still running, and will not delete the Python file in which the DAG is defined, meaning the DAG will appear again in your UI with no history at the next parsing of the `/dags` folder from the scheduler.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,234 @@ tags: ["Database", "SQL", "Components"]
...
+print(req.text)
+```
 
-|F-Key   |Name          |Type                    |Description
-|------- |------------ |---------------------- |--------------
-|        |version_num   |character varying(32)   |*PRIMARY KEY*
+### Example: Retrieve all DAG dependencies
 
-------------------------------------------------------------------------
+Inter DAG dependencies are a powerful tool to get your data orchestration to the next level, including for custom solutions to make sure downstream DAGs get adjusted when changes are made to a DAG with dependencies. These dependencies are visualized in the Airflow UI under **Browse** -> **DAG Dependencies** but currently not accessible through the Airflow REST API.
```
kentdanas, line 193:
```suggestion
Cross-DAG dependencies are a powerful tool to take your data orchestration to the next level. Retrieving data about cross-DAG dependencies from the metadata database can be useful for programmatically implementing custom solutions, for example to make sure downstream DAGs are paused if an upstream DAG is paused. These dependencies can be visualized in the Airflow UI under **Browse** -> **DAG Dependencies**, but are currently not accessible through the Airflow REST API.
```

Updated to use 'cross-dag' language since that's what we use in our other guides (it's not 'better' necessarily, just want to be consistent)

kentdanas (approved):
@TJaniF great job with this! Sooo much better than the old version. I made a few additional comments, but content-wise I think it's good, and I'm approving now. Will loop in @jwitz for his review next :)

guides/airflow-database.md
TJaniF: Yes, I think general works well. (side note: the API docs are really well organized, I found everything I needed immediately)

guides/airflow-database.md
```diff
@@ -8,640 +8,234 @@ tags: ["Database", "SQL", "Components"]
...
-|                                             |  login_count        |integer                       |
-|                                             |  fail\_login\_count   |integer                       |
-|                                             |  created_on         |timestamp without time zone   |
-|                                             |  changed_on         |timestamp without time zone   |
-|  [public.ab\_user.id](#public.ab\_user) |  created\_by\_fk      |integer                       |
-|  [public.ab\_user.id](#public.ab\_user) |  changed\_by\_fk      |integer                       |
+# set location and access information for your airflow instance
+# this example uses the default values for Astro CLI
+ENDPOINT_URL = "http://localhost:8080/"
+user_name = "admin"
```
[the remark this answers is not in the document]
TJaniF: yeah that is a good point, I changed it to use environment variables :)

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+- Postgres
+- MySQL
+- MSSQL
+- SQLite
 
-------------------------------------------------------------------------
+Postgres is by far the most common choice, and is generally recommended by the Airflow community. On Astronomer, each Astro Airflow deployment is equipped with a PostgreSQL database for this purpose. When configuring a custom database backend it is also advisable to make sure [your version is fully supported](https://airflow.apache.org/docs/apache-airflow/stable/howto/set-up-database.html#choosing-database-backend).
 
-## Tables
+Sizing of the metadata database is an important consideration when setting up your Airflow environment for production. Typically, a managed database service will be used so that features like autoscaling and automatic backups can be leveraged. The size you need will depend heavily on the workloads running in your Airflow instance. For reference, when running Airflow locally a 2GB SQLite database will be spun up by default, which is intended for development purposes only. Astro CLI by default will start Airflow with a PostgresSQL database of 1GB.
```
jwitz, line 35:
```suggestion
You should also consider the size of your metadata database when setting up your Airflow environment. Production environments typically use a managed database service, which include features like autoscaling and automatic backups. The size you need will depend heavily on the workloads running in your Airflow instance. For reference, Apache Airflow uses a 2GB SQLite database by default, but this is intended for development purposes only. The Astro CLI starts Airflow environments with a 1GB Postgres database.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-The Airflow metadata database has a total of 30 tables  tables are stored on the public schema by default. The following describe the table structure and reference for the Airflow metadata tables. 
+## Content of the Metadata Database
 
-### public.ab_permission
- 
-**Table Structure:**
+> **Note**: For many use-cases you can access contents from the metadata database via the Airflow UI or the [stable REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html). These points of access are always preferable to querying the metadata database directly!
```
jwitz, line 39:
```suggestion
> **Note**: For many use cases you can access contents from the metadata database via the Airflow UI or the [stable REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html). These points of access are always beat querying the metadata database directly!
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+There are several types of metadata stored in the metadata database.
 
-|  F-Key   | Name  |Type | Description |
-|----------|-------|--------|------------------|
-|          | id    |integer | *PRIMARY KEY*    |
-|          | name  |character varying(100)  | *UNIQUE NOT NULL* |
+- Tables related to users' login information and permissions for a specific airflow instance. The names of these tables all start with `ab_`.
+- Tables storing information used in DAGs, like variables, connections and XComs.
+- Tables storing information about DAG and task runs, which are written to by the scheduler
+- Other tables for example storing DAG code in different formats or information about import errors.
```
jwitz, line 46:
```suggestion
- User login information and permissions. The names of these tables all start with `ab_`.
- Configurations used in DAGs, like variables, connections and XComs.
- Data about DAG and task runs which are generated by the scheduler
- Other minor tables, such as tables which store DAG code in different formats or information about import errors.
```
Is it necessary to list these all as tables? If that's essential, I would move "tables" to your introductory sentence so that you don't have to repeat it with each bulleted item.

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-**Tables referencing this one via Foreign Key Constraints:**
+Changes to the Airflow metadata database configuration and its schema are very common and happen with almost every minor update. For this reason, prior to Airflow 2.3 you could not downgrade your Airflow instance in place. With Airflow 2.3 the `db downgrade` command was added, providing an option to [downgrade Airflow](https://airflow.apache.org/docs/apache-airflow/2.3.0/usage-cli.html#downgrading-airflow).
```
jwitz, line 48:
```suggestion
Changes to the Airflow metadata database configuration and its schema are very common and happen with almost every minor update. For this reason, prior to Airflow 2.3 you should not downgrade your Airflow instance in place. With Airflow 2.3 the `db downgrade` command was added, providing an option to [downgrade Airflow](https://airflow.apache.org/docs/apache-airflow/2.3.0/usage-cli.html#downgrading-airflow).
```

jwitz: Hm this feels a bit out of place here, should this belong in the prior section about configuring a database instance instead?

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
--   [public.ab\_permission\_view](#public.ab\_permission\_view)
+> **Note**: Always backup your database before running any database operations!
 
-------------------------------------------------------------------------
+### Security: Tables related to User Information
```
jwitz, line 52:
```suggestion
### ab_: Tables Related to User Information
```

jwitz: The way we start this section with "Security" makes it sound like that's the name of these tables. I'd prefer to use something closer to the name of the table itself.

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-### public.ab\_permission\_view
+A set of tables starting with `ab_` bundles types of permissions for different sections of Airflow together to form roles which are assigned to individual users. As an admin user, you can access some of the content of these tables in the Airflow UI under the **Security** tab. It is also possible to query and modify [users, roles and permissions](https://airflow.apache.org/docs/apache-airflow/1.10.6/security.html?highlight=ldap#rbac-ui-security) via the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
```
jwitz, line 54:
```suggestion
Tables starting with `ab_` bundle various Airflow permissions into roles which are assigned to individual users. As an admin user, you can access some of the data in these tables in the Airflow UI under the **Security** tab. You can also query and modify [users, roles and permissions](https://airflow.apache.org/docs/apache-airflow/1.10.6/security.html?highlight=ldap#rbac-ui-security) using the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
```

jwitz: Do these tables store just roles, or do they also store each individual user's role? As it's worded now, it sounds like it only stores the roles. If that's not the case, reword to:
```
Tables starting with `ab_` store information about Airflow users, including their permissions to various Airflow features.
```
I think it's okay to abstract some of the details about roles here.

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-|F-Key                                                 |Name            |Type      |Description   |
-|-------------------|---------------|---------|-------------|
-|                                                      |id              |integer   |*PRIMARY KEY*|
-|  [public.ab\_permission.id](#public.ab\_permission)   |permission\_id   |integer   |*UNIQUE\#1*|
-|  [public.ab\_view\_menu.id](#public.ab\_view\_menu)    |view\_menu\_id    |integer   |*UNIQUE\#1*|
+- [Variables](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html)
+- [Connections](https://www.astronomer.io/guides/connections)
+- [XComs](https://www.astronomer.io/guides/airflow-passing-data-between-tasks)
+- [Pools](https://www.astronomer.io/guides/airflow-pools/)
```
jwitz, line 63:
```suggestion
- [Variables](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html).
- [Connections](https://www.astronomer.io/guides/connections).
- [XComs](https://www.astronomer.io/guides/airflow-passing-data-between-tasks).
- [Pools](https://www.astronomer.io/guides/airflow-pools/).
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-**Tables referencing this one via Foreign Key Constraints:**
+The information in these tables can be viewed and modified under the **Admin** tab in the Airflow UI or by using the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).  
 
--   [public.ab\_permission\_view\_role](#public.ab\_permission\_view\_role)
+### Browse: Tables storing Information about DAG and Task Runs
 
-------------------------------------------------------------------------
+Besides storing the metadata mentioned in the previous section, the Airflow metadata database is heavily used by the scheduler to keep track of past and current events. The majority of this data can be found under the **Browse** tab in the Airflow UI or queried via the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
```
jwitz, line 69:
```suggestion
The scheduler depends on the Airflow metadata database to keep track of past and current events. The majority of this data can be found under the **Browse** tab in the Airflow UI or queried via the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
```

guides/airflow-database.md
jwitz: Now that I'm thinking about it more, in the previous note we say that it's always best to find this data via the Airflow UI or REST API. In that case, why are we mentioning the `ab_` table name here while all other tables are not named? Is there some information in these tables that you can get _only_ by querying the db?

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+> **Note**: Always backup your database before running any database operations!
 
-------------------------------------------------------------------------
+### Security: Tables related to User Information
 
-### public.ab\_permission\_view
+A set of tables starting with `ab_` bundles types of permissions for different sections of Airflow together to form roles which are assigned to individual users. As an admin user, you can access some of the content of these tables in the Airflow UI under the **Security** tab. It is also possible to query and modify [users, roles and permissions](https://airflow.apache.org/docs/apache-airflow/1.10.6/security.html?highlight=ldap#rbac-ui-security) via the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
 
-**Table Structure:**
+### Admin: Tables storing Information used in DAGs
```
jwitz, line 56:
```suggestion
### DAG Values
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+- [Variables](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html)
+- [Connections](https://www.astronomer.io/guides/connections)
+- [XComs](https://www.astronomer.io/guides/airflow-passing-data-between-tasks)
+- [Pools](https://www.astronomer.io/guides/airflow-pools/)
 
-**Tables referencing this one via Foreign Key Constraints:**
+The information in these tables can be viewed and modified under the **Admin** tab in the Airflow UI or by using the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).  
 
--   [public.ab\_permission\_view\_role](#public.ab\_permission\_view\_role)
+### Browse: Tables storing Information about DAG and Task Runs
```
jwitz, line 67:
```suggestion
### DAG and Task Run Metadata
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-### public.ab\_permission\_view\_role
+- **DAG Runs** stores information on all past and current Dag Runs including if they were successful, scheduled or manually triggered and detailed timing information
+- **Jobs** contains data used by the scheduler to store information about past and current jobs of different types (`SchedulerJob`, `TriggererJob`, `LocalTaskJob`)
+- **Audit logs** shows events of various types that were logged to the metadata database (e.g. DAGs being paused or unpaused or tasks being run)
+- **Task Instances** contains a record of every task instance (a task with a associated timestamp when it ran) with a variety of attributes such as the priority weight, duration or the URL to the task log
+- **Task Reschedule** lists tasks that have been rescheduled
+- **Triggers** shows all currently running [triggers](https://www.astronomer.io/guides/deferrable-operators)
+- **SLA Misses** keeps track of tasks that missed their [SLA](https://www.astronomer.io/guides/error-notifications-in-airflow/#airflow-slas).
```
jwitz, line 77:
```suggestion
- **DAG Runs** stores information on all past and current DAG runs including whether they were successful, whether they were scheduled or manually triggered, and detailed timing information.
- **Jobs** contains data used by the scheduler to store information about past and current jobs of different types (`SchedulerJob`, `TriggererJob`, `LocalTaskJob`).
- **Audit logs** shows events of various types that were logged to the metadata database (for example, DAGs being paused or tasks being run).
- **Task Instances** contains a record of every task run with a variety of attributes such as the priority weight, duration, or the URL to the task log.
- **Task Reschedule** lists tasks that have been rescheduled.
- **Triggers** shows all currently running [triggers](https://www.astronomer.io/guides/deferrable-operators).
- **SLA Misses** keeps track of tasks that missed their [SLA](https://www.astronomer.io/guides/error-notifications-in-airflow/#airflow-slas).
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-|F-Key  |Name |Type |Description|
-|------|----|----|----------|
-|       |id   |integer |   *PRIMARY KEY*|
-|[public.ab\_permission\_view.id](#public.ab\_permission\_view)   |permission\_view\_id   |integer   |*UNIQUE\#1*|
-|[public.ab\_role.id](#public.ab\_role)                         |role\_id              |integer   |*UNIQUE\#1*|
+- **Browse** -> **DAG Dependencies** shows a visual representation of DAG Dependencies
+- The source code of DAGs can be found by clicking on a DAG name from the main view and then navigating to the **Code** view
+- Import errors will show up on top of your list of dags if they occur
+- DAG tags will show up underneath their respective DAG with a cyan background.
```
jwitz, line 86:
```suggestion
- **Browse** -> **DAG Dependencies** shows a visual representation of DAG Dependencies.
- The source code of DAGs can be found by clicking on a DAG name from the main view and then going to the **Code** view.
- Import errors appear at the top of the **DAGs** view in the UI.
- DAG tags will appear underneath their respective DAG with a cyan background.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+Most of the information in these tables is accessible via the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
```
jwitz, line 88: When we link out to the REST API docs in each of these sections, can we link to a specific query/ doc that's related to the info in the database? If not, let's just mention that you can access most of _all_ of this data from the REST API in the introductory section.

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-### public.ab\_register\_user
+## Airflow Metadata Database Best Practices
 
-**Table Structure:**
+1. When up- or downgrading Airflow, which commonly includes changes to your metadata database, make sure you always have a backup of your metadata database and follow the [recommended steps for changing Airflow versions](https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading.html?highlight=upgrade): backup the database, check for deprecated features, pause all DAGs, and make sure no tasks are running.
```
jwitz, line 92:
```suggestion
1. When upgrading or downgrading Airflow, always follow the [recommended steps for changing Airflow versions](https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading.html?highlight=upgrade): back up the metadata database, check for deprecated features, pause all DAGs, and make sure no tasks are running.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+2. Use caution when [pruning old records](https://airflow.apache.org/docs/apache-airflow/stable/usage-cli.html#purge-history-from-metadata-database) from your database with `db clean`. The `db clean` command allows you to delete records older than `--clean-before-timestamp` from all metadata database tables or a list of tables specified.
```
jwitz, line 94:
```suggestion
2. Use caution when [pruning old records](https://airflow.apache.org/docs/apache-airflow/stable/usage-cli.html#purge-history-from-metadata-database) from your database with `db clean`. The `db clean` command allows you to delete records older than `--clean-before-timestamp` from all metadata database tables or a list of tables specified.
```
How does someone practice caution with pruning? What specifically should they avoid, or what are the potential negative effects?

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
-||          id                  |integer                       |*PRIMARY KEY*
-||          first_name          |character varying(64)         |*NOT NULL*
-||          last_name           |character varying(64)         |*NOT NULL*
-||          username            |character varying(64)         |*UNIQUE NOT NULL*
-||          password            |character varying(256)        |
-||          email               |character varying(64)         |*NOT NULL*
-||          registration_date   |timestamp without time zone   |
-||          registration_hash   |character varying(256)        |
+3. Keep in mind that every time you access data from the metadata database from within a DAG (for example by fetching a variable, pulling from XCom or using a connection ID) you will make a connection to the metadata database which requires compute resources. It is therefore best practice to try to keep these actions within tasks, which will only create a connection to the database at run time. If they are written as top level code, connections will be created every time the scheduler parses the DAG file (every 30 seconds by default!).
```
jwitz, line 96:
```suggestion
3. Accessing the metadata database from within a DAG (for example by fetching a variable, pulling from XCom, or using a connection ID) requires compute resources. It is therefore best practice to keep these actions within tasks, which creates a connection to the database only for the run time of the task. If these connections are written as top level code, connections are created every time the scheduler parses the DAG file, which is every 30 seconds by default!
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+4. Memory in the Airflow metadata database can be limited depending on your setup and running low on memory in your metadata database can cause performance issues in Airflow. This is one of the many reasons why we highly advise against moving large amounts of data via XCom, and recommend using a cleanup and archiving mechanism in any production deployments.
```
jwitz, line 98:
```suggestion
4. Memory in the Airflow metadata database can be limited depending on your setup, and running low on memory in your metadata database can cause performance issues in Airflow. This is one of the many reasons why we highly advise against moving large amounts of data via XCom, and recommend using a cleanup and archiving mechanism in any production deployments.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-### public.ab_role
+5. Since the metadata database is critical for the scalability and resiliency of your Airflow deployment, it is best practice at least for production Airflow deployments to use a managed database service, for example [AWS RDS](https://aws.amazon.com/rds/) or [Google Cloud SQL](https://cloud.google.com/sql).
```
jwitz, line 100:
```suggestion
5. Since the metadata database is critical for the scalability and resiliency of your Airflow deployment, it is best practice to use a managed database service for production environments, for example [AWS RDS](https://aws.amazon.com/rds/) or [Google Cloud SQL](https://cloud.google.com/sql).
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-**Table Structure:**
+## Examples Methods for Accessing Data of Interest
```
jwitz, line 102:
```suggestion
## Examples: Programmatically Access the Metadata Database
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+The preferred methods for retrieving data from the metadata database are using the Airflow UI or making a GET request to the [stable Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html). Between the UI and API large parts of the metadata database can be viewed without the risk of accidentally making harmful changes to the database that direct querying would pose. For use cases where neither the Airflow UI nor the REST API can provide sufficient data it is recommended to use SQLAlchemy to query the metadata database in order to have an additional layer of abstraction on top of the tables, which makes your code less sensitive to minor schema changes.
```
jwitz, line 104:
```suggestion
When possible, the best methods for retrieving data from the metadata database are using the Airflow UI or making a GET request to the [stable Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html). Between the UI and API, much of the metadata database can be viewed without the risk inherent in direct querying. For use cases where neither the Airflow UI nor the REST API can provide sufficient data, we recommended using SQLAlchemy to query the metadata database. This provides an additional layer of abstraction on top of the tables, which makes your code less sensitive to minor schema changes.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+In the Airflow UI you can navigate to **Browse** -> **Task Instances** and filter the task instances for all with a state of `success`. The `Record Count` will be on the right side of your screen.
 
-### public.ab_user
+![Count successful tasks Airflow UI](<https://assets2.astronomer.io/main/guides/your-guide-folder/successful_tasks_UI.png>)
 
-**Table Structure:**
+Using the [stable REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#section/Overview) to query the metadata database is an equally valid way to get information. Make sure you have [correctly authorized API use](https://airflow.apache.org/docs/apache-airflow/stable/security/api.html) in your Airflow instance and set the `ENDPOINT_URL` to the correct location (for local development: `http://localhost:8080/`).
 
-------------------------------------------------------------------------
+```python
```
jwitz, line 118: What kind of code is this? Just a standard Python script? It would be helpful to introduce this snippet a little more

guides/airflow-database.md
jwitz: We already go into pretty thorough detail about accessing this data via the UI earlier in the guide. How do you feel about deleting the UI information here and instead focusing on API and direct queries instead?

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
-|                                         |show_datatable    |boolean                    |
-|                                         |show_sql          |boolean                    |
-|                                         |height            |integer                    |
-|                                         |default_params    |character varying(5000)    |
-|                                         |x\_is\_date         |boolean                    |
-|                                         |iteration_no      |integer                    |
-|                                         |last_modified     |timestamp with time zone   |
+Cross-DAG dependencies are a powerful tool to take your data orchestration to the next level. Retrieving data about cross-DAG dependencies from the metadata database can be useful for programmatically implementing custom solutions, for example to make sure downstream DAGs are paused if an upstream DAG is paused. These dependencies can be visualized in the Airflow UI under **Browse** -> **DAG Dependencies**, but are currently not accessible through the Airflow REST API.
```
jwitz, line 202:
```suggestion
Cross-DAG dependencies are a powerful tool to take your data orchestration to the next level. Retrieving data about cross-DAG dependencies from the metadata database can be useful for programmatically implementing custom solutions, for example to make sure downstream DAGs are paused if an upstream DAG is paused. These dependencies can be visualized in the Airflow UI under **Browse** -> **DAG Dependencies**, but they are not accessible through the Airflow REST API.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+In situations like these it is possible to use SQLAlchemy with [Airflow models](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/index.html) to access data from the metadata database similar to how the webserver interacts with it. Querying the metadata database in this way is preferred over direct SQL statements because it adds an additional layer of abstraction that makes your code less sensitive to minor schema changes. If you are running Airflow in a dockerized setting you will have to run the script below from within your scheduler container.
```
jwitz, line 204:
```suggestion
To programmatically access this information, you can use SQLAlchemy with [Airflow models](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/index.html) to access data from the metadata database. Note that if you are running Airflow in a Dockerized setting, you have to run the script below from within your scheduler container.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+with Session(engine) as session:
+    result = session.query(SerializedDagModel).first()
+    print(result.get_dag_dependencies())
+```
 
-### public.dag
+### Example: Retrieving Alembic Version
 
-**Table Structure:**
+In very rare cases you may need a value from the metadata database that is not accessible in the Airflow UI, via the Airflow REST API or by using a SQLAlchemy model. In this case you can query the metadata database directly, while keeping in mind that every Airflow update may change the schema and therefore break your code and that you can corrupt your Airflow instance by directly manipulating the metadata database.
```
jwitz, line 227:
```suggestion
In very rare cases, you might need a value from the metadata database which is not accessible through any of the methods we've discussed. In this case, you can query the metadata database directly. Before you do so, that you can corrupt your Airflow instance by directly manipulating the metadata database, especially if the schema changes between upgrades.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+The query below retrieves the current [alembic version id](https://alembic.sqlalchemy.org/en/latest/), which is not accessible through any of the recommended ways of interacting with the metadata database.
```
jwitz, line 229: Can we explain why you'd want to query this within the context of Airflow?

jwitz (changes requested):
You're on a roll @TJaniF! Just some minor comments and questions throughout, but requesting changes on some aspects of the examples at the end of the guide and some of the titles.

guides/airflow-database.md
TJaniF: No, the "tables" bit is kinda self-evident. I am not sure why I wrote it this way, yours is much cleaner.

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
 
 ## Overview
 
-> Note: This guide has been written and tested with Airflow Airflow 1.10.10. New tables may be introduced in newer versions. 
+The metadata database is a core component of Airflow, and is used to store crucial information about both the configuration of your Airflow environment as well as all metadata relevant for the scheduler regarding past and present DAG and task runs.
```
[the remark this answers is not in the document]
TJaniF: I realized I had worded this ambiguously, because the Airflow configuration is _not_ stored in the metadata database (rather in the airflow.cfg file). I changed it to:

> The metadata database is a core component of Airflow. It stores crucial information like the configuration of your Airflow environment's roles and permissions, as well as all metadata for past and present DAG and task runs.

guides/airflow-database.md
```diff
@@ -8,640 +8,244 @@ tags: ["Database", "SQL", "Components"]
...
+In this guide, we will explain everything you need to know about the Airflow metadata database to ensure a healthy Airflow environment, including:
 
-Even though we don't recommend modifying the values directly on the database, as doing so might affect dependencies, understanding the underlying structure can be very useful when it comes to building your own reports or querying directly against the database. You can find some useful queries that go directly against these tables in our [Useful SQL queries for Apache Airflow](https://www.astronomer.io/guides/airflow-queries/) guide. 
+- Specifications for the database instance, including type and sizing considerations
+- Important content stored in the database
+- Key best practices for interacting with the metadata database
+- Different methods for accessing data of interest
```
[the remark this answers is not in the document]
TJaniF: Since we do not really cover how to configure a database instance (only explain on a very high level that you can and some of the options) I changed the first bullet point to "Database specifications."

guides/airflow-database.md
TJaniF: That paragraph is in this section because a very early draft of this included schemas of some tables and then the note was a warning that the schemas might change any time. But now without any schemas in the text it makes more sense in the previous section, yes.

guides/airflow-database.md
TJaniF: >  Is there some information in these tables that you can get only by querying the db?

I spent several hours when writing the guide to find an answer to that question so we had anything to query for the examples 😂 . The answer is that the API is too good and barely anything in the metadata database is not accessible over it or the UI. 

There is some info in the `ab_` tables that is only accessible directly like registered users (if someone has their Airflow set up for self-registration, idk why you'd want to do that). But nothing really useful. I agree with dropping the info about the shared table name prefix because even if you'd directly query you'd have to look up the exact table name anyways. 

I used "Security: ...", "Admin: ....", "Browse: ...." because those are the headers in the UI under which to find the information. Maybe it works if I make this more explicit as "UI Tab Security: ....." ? My goal with it is to link the content to the UI structure, which is probably pretty familiar for the DAG authors reading the guide.

guides/airflow-database.md
TJaniF: See my comment above.

guides/airflow-database.md
TJaniF: See my comment above.

guides/airflow-database.md
TJaniF: The API docs are structured in a different way so we could link to an example like "this is how to find import errors with the API" but not to any header that has all info for the section. I think putting the link in the intro section is cleaner.

guides/airflow-database.md
TJaniF: Hmmm... I like the title change, I definitely think programatically should be the focus of this section but I am unsure about deleting the UI info because I like the flow of presenting a common use case like successfully completed tasks and then show the two options (I don't know if all DAG authors know that you can filter Task Instances in the UI and a count is shown and I think the DAG dependencies feature in Browse is also not that well known).

For the first example I moved the note on how to access via UI to the bottom of the section, highlighting the API way, which people will most likely prefer for this and I removed "how to find DAG dependencies" in the UI from the "Other Tables" section above, to make it less repetitive. 

Is that a compromise that works for you? 🙂

guides/airflow-database.md
TJaniF: Yes, a Python script because I prefer this way of API calls over using Bash. I added an explanation at the start of it.

guides/airflow-database.md
TJaniF: The truth is that I used the alembic version as an example because it was one of the only things I could find that can't be queried via either the API or the preferred SQLAlchemy route. I asked a DB admin friend and they said they needed that version before when doing complex database migrations but that is as far as my alembic knowledge goes 😅 . I added that as explanation. 

I decided to keep the example in so we can show how one can still directly query the db so people who need this for something quick and dirty and don't want to search through the airflow.models have a code snippet for direct queries. This will be more intuitive to use for many (me included) than the recommended airflow.models via SQLAlchemy.

guides/airflow-database.md
jwitz: Updated my suggestion to include a link to the current Airflow docs

guides/airflow-database.md
```diff
@@ -8,640 +8,251 @@ tags: ["Database", "SQL", "Components"]
...
-|------- |-------------------|-----------------------------|------------------|
-||          id                  |integer                       |*PRIMARY KEY*
-||          first_name          |character varying(64)         |*NOT NULL*
-||          last_name           |character varying(64)         |*NOT NULL*
-||          username            |character varying(64)         |*UNIQUE NOT NULL*
-||          password            |character varying(256)        |
-||          email               |character varying(64)         |*NOT NULL*
-||          registration_date   |timestamp without time zone   |
-||          registration_hash   |character varying(256)        |
+2. Use caution when [pruning old records](https://airflow.apache.org/docs/apache-airflow/stable/usage-cli.html#purge-history-from-metadata-database) from your database with `db clean`, avoid pruning records that future runs might depend on e.g. via the `depends_on_past` argument. The `db clean` command allows you to delete records older than `--clean-before-timestamp` from all metadata database tables or a list of tables specified.
```
jwitz, line 95:
```suggestion
2. Use caution when [pruning old records](https://airflow.apache.org/docs/apache-airflow/stable/usage-cli.html#purge-history-from-metadata-database) from your database with `db clean`. For example, avoid pruning records could affect future runs for tasks that use the `depends_on_past` argument. The `db clean` command allows you to delete records older than `--clean-before-timestamp` from all metadata database tables or a list of tables specified.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,251 @@ tags: ["Database", "SQL", "Components"]
...
+```
 
-**Indexes:**
+### Example: Retrieving Alembic Version
 
--   **idx\_root\_dag\_id** root\_dag\_id
+In very rare cases, you might need a value from the metadata database which is not accessible through any of the methods we've discussed. In this case, you can query the metadata database directly. Before you do so, remember that you can corrupt your Airflow instance by directly manipulating the metadata database, especially if the schema changes between upgrades.
 
-------------------------------------------------------------------------
+The query below retrieves the current [alembic version id](https://alembic.sqlalchemy.org/en/latest/), which is not accessible through any of the recommended ways of interacting with the metadata database. Database administrators might need the version id for complex data migration operations.
```
jwitz, line 236:
```suggestion
The query below retrieves the current [alembic version ID](https://alembic.sqlalchemy.org/en/latest/), which is not accessible through any of the recommended ways of interacting with the metadata database. Database administrators might need the version ID for complex data migration operations.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,251 @@ tags: ["Database", "SQL", "Components"]
 
 ## Overview
 
-> Note: This guide has been written and tested with Airflow Airflow 1.10.10. New tables may be introduced in newer versions. 
+The metadata database is a core component of Airflow. It stores crucial information like the configuration of your Airflow environment's roles and permissions, as well as all metadata for past and present DAG and task runs.
 
-We at Astronomer often get asked about the structure of the underlying Airflow metadata database. The Airflow metadata database stores configurations, such as variables and connections, user information, roles, and policies. It is also the Airflow Scheduler's source of truth for all metadata regarding DAGs, schedule intervals, statistics from each run, and tasks. 
+A healthy metadata database is critical for your Airflow environment. Losing data stored in the metadata database can can both interfere with running DAGs and prevent you from accessing data for past DAG runs. As with any core Airflow component, having a backup and disaster recovery plan in place for the metadata database is essential.
```
jwitz, line 13:
```suggestion
A healthy metadata database is critical for your Airflow environment. Losing data stored in the metadata database can both interfere with running DAGs and prevent you from accessing data for past DAG runs. As with any core Airflow component, having a backup and disaster recovery plan in place for the metadata database is essential.
```

guides/airflow-database.md
```diff
@@ -8,640 +8,251 @@ tags: ["Database", "SQL", "Components"]
...
+- User login information and permissions. The names of these tables all start with `ab_`.
+- Information used in DAGs, like variables, connections and XComs.
+- Data about DAG and task runs which are generated by the scheduler
+- Other minor tables, such as tables which store DAG code in different formats or information about import errors.
 
-### public.ab\_permission\_view
+Most of this data can be queried using the [Airflow REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html).
 
-**Table Structure:**
+### UI Tab Security: Tables related to User Information
```
jwitz, line 56:
```suggestion
### Security (User Information)
```
Maybe something like this? Don't want to make the titles too wordy

guides/airflow-database.md
```diff
@@ -8,640 +8,251 @@ tags: ["Database", "SQL", "Components"]
...
+- [Variables](https://airflow.apache.org/docs/apache-airflow/stable/howto/variable.html).
+- [Connections](https://www.astronomer.io/guides/connections).
+- [XComs](https://www.astronomer.io/guides/airflow-passing-data-between-tasks).
+- [Pools](https://www.astronomer.io/guides/airflow-pools/).
 
-------------------------------------------------------------------------
+The information in these tables can be viewed and modified under the **Admin** tab in the Airflow UI.
 
-### public.ab\_permission\_view\_role
+### UI Tab Browse: Tables storing Information about DAG and Task Runs
```
jwitz, line 71:
```suggestion
### Browse (DAG and Task Runs)
```

guides/airflow-database.md
```diff
@@ -8,640 +8,251 @@ tags: ["Database", "SQL", "Components"]
...
 
-------------------------------------------------------------------------
+A set of tables store information about Airflow users, including their [permissions](https://airflow.apache.org/docs/apache-airflow/1.10.6/security.html?highlight=ldap#rbac-ui-security) to various Airflow features. As an admin user, you can access some of the content of these tables in the Airflow UI under the **Security** tab.
 
-|F-Key                                                 |Name            |Type      |Description   |
-|-------------------|---------------|---------|-------------|
-|                                                      |id              |integer   |*PRIMARY KEY*|
-|  [public.ab\_permission.id](#public.ab\_permission)   |permission\_id   |integer   |*UNIQUE\#1*|
-|  [public.ab\_view\_menu.id](#public.ab\_view\_menu)    |view\_menu\_id    |integer   |*UNIQUE\#1*|
+### UI Tab Admin: Tables storing Information used in DAGs
```
jwitz, line 60:
```suggestion
### Admin (DAG Configurations and Variables)
```

TJaniF:
Thank you for the second round of edits! :) (and catching that I linked to a 1.10 doc 🤦 )

> Actually one final comment @TJaniF : I think you could do away with the "UI" parts of your headers in the database content section. I understand your point about DAG authors being familiar with those UI elements. But the overall topic is "Database Content", not "Where to find database elements in the UI". Because of that, the UI framing might throw some people off who are either not familiar with Airflow or are approaching this guide from a different perspective than we're anticipating.
> 
> I don't feel super strongly either way though, so I trust you to do what feels right for the titles!

I think it makes sense to change the emphasis to the db content and have the UI tab name in parenthesis, so instead of "Security (User Information)" -> "User Information (Security)". So the structure is there for those who are familiar with the UI to make the connection that they've been looking at the database the whole time without realizing (that was my experience when researching the guide 😄 ) but it should not throw off people who don't use the UI.

PR #389, opened 2022-06-28
Add decorators section to Task Groups guide

kentdanas:
Title says it all. Adding a section with an example DAG we used in a recent decorators webinar on how to define task groups using decorators. @jwitz let me know if you have any comments.

### DIFF
--- a/guides/task-groups.md
+++ b/guides/task-groups.md
@@ -57,9 +57,9 @@ In the Airflow UI, Task Groups look like tasks with blue shading. When we expand
 
 ## Using the Task Group Decorator
 
-Another way of defining Task Groups in your DAGs is by using the Task Group decorator. Note that this feature requires Airflow 2.1+. The Task Group decorator works like other [decorators in Airflow](https://www.astronomer.io/guides/airflow-decorators) by allowing you to define your Task Group with less boilerplate code using the TaskFlow API. The functionality of Task Groups does not change when defining them with a decorator, but for DAG authors who use the `@task` Python task decorator in their DAGs, it may be a more stylistically consistent option.
+Another way of defining Task Groups in your DAGs is by using the Task Group decorator. Note that this feature requires Airflow 2.1+. The Task Group decorator works like other [decorators in Airflow](https://www.astronomer.io/guides/airflow-decorators) by allowing you to define your Task Group with less boilerplate code using the TaskFlow API. Using Task Group decorators doesn't change the functionality of Task Groups, but if you already use task decorators in your DAGs then they can make your code formatting more consistent.
 
-At a high level, to use the decorator you must use `@task_group` before a Python function which calls the functions of tasks that should go in the Task Group. For example:
+To use the decorator, add `@task_group` before a Python function which calls the functions of tasks that should go in the Task Group. For example:
 
 ```python
 @task_group(group_id="tasks")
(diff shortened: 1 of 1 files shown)


guides/task-groups.md
```diff
@@ -55,6 +55,89 @@ In the Airflow UI, Task Groups look like tasks with blue shading. When we expand
 
 > **Note:** When your task is within a Task Group, your callable `task_id` will be the `task_id` prefixed with the `group_id` (i.e. `group_id.task_id`). This ensures the uniqueness of the task_id across the DAG. This is important to remember when calling specific tasks with XCOM passing or branching operator decisions.
 
+## Using the Task Group Decorator
+
+Another way of defining Task Groups in your DAGs is by using the Task Group decorator. Note that this feature requires Airflow 2.1+. The Task Group decorator works like other [decorators in Airflow](https://www.astronomer.io/guides/airflow-decorators) by allowing you to define your Task Group with less boilerplate code using the TaskFlow API. The functionality of Task Groups does not change when defining them with a decorator, but for DAG authors who use the `@task` Python task decorator in their DAGs, it may be a more stylistically consistent option.
+
+At a high level, to use the decorator you must use `@task_group` before a Python function which calls the functions of tasks that should go in the Task Group. For example:
```
jwitz, line 62:
```suggestion
To use the decorator, add `@task_group` before a Python function which calls the functions of tasks that should go in the Task Group. For example:
```

guides/task-groups.md
```diff
@@ -55,6 +55,89 @@ In the Airflow UI, Task Groups look like tasks with blue shading. When we expand
...
+
+```python
+@task_group(group_id="tasks")
+def my_independent_tasks():
+    task_a()
+    task_b()
+    task_c()
+```
+
+This function will create a Task Group with three independent tasks that should be defined elsewhere in the DAG.
```
jwitz, line 72:
```suggestion
This function creates a Task Group with three independent tasks that are defined elsewhere in the DAG.
```

guides/task-groups.md
```diff
@@ -55,6 +55,89 @@ In the Airflow UI, Task Groups look like tasks with blue shading. When we expand
...
+
+    
+task_group_example = task_group_example()
+```
+
+The resulting DAG looks like this:
+
+![Decorated Task Group](https://assets2.astronomer.io/main/guides/task-groups/decorated_task_group.png)
+
+There are a few things to note about this DAG and when using the Task Group decorator:
```
jwitz, line 135:
```suggestion
There are a few things to consider when using the Task Group decorator:
```

guides/task-groups.md
```diff
@@ -55,6 +55,89 @@ In the Airflow UI, Task Groups look like tasks with blue shading. When we expand
...
+
+- If downstream tasks require the output of tasks that are in the Task Group, then the Task Group function must return a result. In this example, we return a dictionary with two values, one from each of the tasks in the Task Group, that are then passed to the downstream `load()` task.
```
jwitz, line 137:
```suggestion
- If downstream tasks require the output of tasks that are in the Task Group decorator, then the Task Group function must return a result. In the previous example, we return a dictionary with two values, one from each of the tasks in the Task Group, that are then passed to the downstream `load()` task.
```

guides/task-groups.md
```diff
@@ -55,6 +55,89 @@ In the Airflow UI, Task Groups look like tasks with blue shading. When we expand
...
+- If your Task Group function does not return any output, you will need to use bitshift operators (`<<` or `>>`) to define dependencies, rather than calling the function with the TaskFlow API.
```
jwitz, line 138:
```suggestion
- If your Task Group function returns an output, you can simply call the function from your DAG with the TaskFlow API like in the previous example. If your Task Group function does not return any output, you must use bitshift operators (`<<` or `>>`) to define dependencies to the Task Group like you would with a standard task.
```

jwitz (approved):
V minor wording suggestions, otherwise lgtm!

PR #390, opened 2022-06-30
moving the s3hook import to the methods per the 2-3-2 break that happens

erdos2n:
Airflow 2.3.2, with custom xcoms backend, has an cyclical import issue. To avoid this from happening, the S3Hook import needs to be added to the methods in which it is used. 

I have updated the code blocks to reflect this fix.

### DIFF
--- a/guides/custom-xcom-backends.md
+++ b/guides/custom-xcom-backends.md
@@ -85,7 +85,6 @@ For this example, we'll define a Python file with serialization/deserialization 
 ```python
 from typing import Any
 from airflow.models.xcom import BaseXCom
-from airflow.providers.amazon.aws.hooks.s3 import S3Hook
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


kentdanas:
@erdos2n thanks for bringing this up. Do you think this issue will be fixed in a future Airflow release? If so, I'm wondering if it might make sense to just add a note about it, since not all readers will be on 2.3.2 specifically

kentdanas:
@erdos2n FYI, we've frozen updates to this repo for now until we migrate guides to their new home on our docs site (should happen in the next couple of weeks) so I'm going to close this PR. Our XCom backends guide is slated for an update anyway, so we'll take a look at this then. Feel free to submit an issue in our docs repo if you notice other issues: https://github.com/astronomer/docs/issues

PR #395, opened 2022-07-15
Update scheduling-in-airflow.md

magdagultekin:
Example used in the section about timetables mentions 16:30, not 14:30.


kentdanas (approved):
Thanks for catching @magdagultekin!

PR #397, opened 2022-07-15
Update airflow-scaling-workers.md

RNHTTR:
`parallelism` is apparently max number of queued/running tasks per scheduler. TIL!

See: https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#parallelism

### DIFF
--- a/guides/airflow-scaling-workers.md
+++ b/guides/airflow-scaling-workers.md
@@ -39,7 +39,7 @@ You should modify environment-level settings if you want to tune performance acr
 
 Core settings control the number of processes running concurrently and how long those processes can run. The associated environment variables for all parameters in this section are formatted as `AIRFLOW__CORE__PARAMETER_NAME`.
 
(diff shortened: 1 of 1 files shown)


kentdanas:
Ah, good catch on the per scheduler distinction @RNHTTR! 

My only question with this update is the change from "queued" to "scheduled" for the task state users will see if max parallelism has been reached. My understanding was that tasks would remain in a queued state until there is an available slot for them, at which point they'll be moved to scheduled - if this is the case then the original language would be correct. Can you double check on this piece?

RNHTTR:
Tasks in both the `queued` and `running` states count against `parallelism`. I think it'd be somewhat difficult to demo this, as tasks generally move quickly from `queued` to `running`, although I can give it a try.

A [task's lifecycle](https://airflow.apache.org/docs/apache-airflow/stable/concepts/tasks.html#task-instances) typically goes...

```
None --> scheduled --> queued --> running --> success/up-for-retry/failed
```

Tasks _can_ get stuck in `queued`, but IMO it's unlikely that increasing `parallelism` would help. On the other hand, increasing `parallelism` would likely help tasks stuck in `scheduled`. Take for example a bunch of long-running tasks taking up all of the `parallelism` slots. The scheduler knows that tasks need to be scheduled, but there are no `parallelism` "slots" remaining, because a bunch of tasks are still running.

kentdanas:
Gotcha, I just had it backwards in my head then :)

Approved the PR, you should be able to merge now!

kentdanas (approved):

PR #398, opened 2022-07-19
Remove note about termination grace period on astronomer

kentdanas:
This guide needs more updates, but for now I'm just removing some erroneous notes about features on Astronomer.

### DIFF
--- a/guides/airflow-executors-explained.md
+++ b/guides/airflow-executors-explained.md
@@ -21,11 +21,11 @@ We'll give the Sequential Executor an honorable mention, too.
 
 Once a DAG is defined (perhaps with the help of an _Operator_), the following needs to happen in order for a single or set of "tasks" within that DAG to execute and be completed from start to finish:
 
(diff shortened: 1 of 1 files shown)


guides/airflow-executors-explained.md
```diff
@@ -21,19 +21,19 @@ We'll give the Sequential Executor an honorable mention, too.
 
 Once a DAG is defined (perhaps with the help of an _Operator_), the following needs to happen in order for a single or set of "tasks" within that DAG to execute and be completed from start to finish:
```
TJaniF, line 22:
```suggestion
Once a DAG is defined the following needs to happen in order for the tasks within that DAG to execute and be completed from start to finish:
```

kentdanas: Thanks for the correction! Yeah this one needs to be rewritten in general, but we'll wait to do that until the new executor is released :)

PR #401, opened 2022-07-20
Deprecate legacy guides

kentdanas:
Removing two legacy guides that we no longer want to have published:

- Intro to Kubernetes: because there are better and more up to date k8s 101 resources elsewhere, and it isn't directly related to running Airflow
- Kerberos: because it is out of date, not widely viewed, doesn't work for our customers, and doesn't add any value on top of OSS Airflow docs


kentdanas:
@allenbrady-astro can we get redirects from these two back to the main guides site?

TJaniF (approved):

allenbrady-astro:
Hey @kentdanas, the redirects are now live so should be safe to remove these guides now! 👍

PR #407, opened 2022-07-28
Update module imports in Orchestrating Snowflake guide

josh-fell:
A few module imports were using either deprecated, legacy, or non-ideal paths. This PR updates the code snippets in the "Orchestrating Snowflake Queries with Airflow" guide to reflect module import paths that should be used in our examples.

### DIFF
--- a/guides/airflow-snowflake.md
+++ b/guides/airflow-snowflake.md
@@ -19,11 +19,11 @@ In this guide, we cover everything you need to know to make the most out of Airf
 
 ## Using Snowflake Providers
 
-There are multiple open source packages that you can use to orchestrate Snowflake in Airflow: 
+There are multiple open source packages that you can use to orchestrate Snowflake in Airflow:
 
 - The [Snowflake provider package](https://registry.astronomer.io/providers/snowflake) contains hooks, operators, and transfer operators for Snowflake maintained by the Airflow community.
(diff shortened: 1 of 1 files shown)


guides/airflow-snowflake.md
```diff
@@ -266,14 +266,13 @@ def transform_dataframe(df: DataFrame):
     return purchase_dates
 
 
-dag = DAG(
+with DAG(
```
josh-fell, line 269: Updated to use the DAG context manager approach to be consistent within the guide.

guides/airflow-snowflake.md
```diff
@@ -39,48 +39,48 @@ Modules for orchestrating basic queries and functions in Snowflake include:
...
 
 ```python
 import json
+from pendulum import datetime
 
 from pathlib import Path
 
 from airflow import DAG
 from airflow.models.baseoperator import chain
-from airflow.operators.dummy_operator import DummyOperator
```
josh-fell, line 79: This path is deprecated and with Airflow 2.3.0, `DummyOperator` was replaced with `EmptyOperator` for more inclusive language.

kentdanas: Dang, I thought I had caught this everywhere. Thanks for the update!

kentdanas (approved):

PR #413, opened 2022-08-08
Update dynamically-generating-dags.md

magdagultekin:
I followed the multiple-file method to generate DAGs dynamically and `dag_id`, `scheduletoreplace`, and `querytoreplace` don't get replaced by the values from JSON files. 

I've tested it with the proposed code and everything looks good, below screen shot of `dag_file_1`: 
image


kentdanas (approved):
Thanks @magdagultekin! This one hasn't been updated in a while, appreciate the bug fixes :)

PR #421, opened 2022-08-10
Astro Python SDK updates

kentdanas:
Updating our two guides that have Astro Python SDK examples to 1) bring the language in line with current project naming, and 2) change the example code to work with v 1.0

@jwitz hoping to have these published by 8/23, so we have some time.


guides/astro-python-sdk-etl.md
```diff
@@ -0,0 +1,280 @@
...
+
+- The `load_file` and `append` functions take care of loading our raw data from S3 and appending data to our reporting table respectively. We didn't have to write any extra code to get the data into Snowflake. In this implementation we also have a `load_file` task for each file instead of one task for all files in S3, which supports atomicity.
+- Using the `transform` function, we easily executed SQL to combine our data from multiple tables. The results are automatically stored in a table in Snowflake. We didn't have to use the `SnowflakeHook` or write any of the code to execute the query.
+- We seamlessly transitioned to a transformation in Python with the `df` function without needing to explicitly convert the results of our previous task to a Pandas dataframe. We then wrote the output of our transformation to our aggregated reporting table in Snowflake using the `target_table` parameter, so we didn't have to worry about storing the data in XCom.
+- We only had to define our connection in tasks with no upstream (`load_file` and `create_reporting_table`). All downstream task that inherit from a task with a connection defined will use it.
```
jwitz, line 278:
```suggestion
- We didn't have to redefine our Airflow connections in any tasks that were downstream of our original definitions (such as `load_file` and `create_reporting_table`). Any downstream task that inherits from a task with a defined connection can use the same connection without additional configuration.
```

PR #426, opened 2022-08-19
2.4 Update to Dynamic Tasks

TJaniF:
Added a section on mapping over multiple parameters with the three methods:
- cross product
- expand kwargs
- zip

One of the examples for expanding kwargs is not implemented yet, but TP said he is planning on doing it. 🙂

I showed what zip does in comments in the code example, am on the fence on adding a small zip(["a","b","c"], [1,2,3]) -> [("a",1),("b",2),("c",3)] into the text. Might be a bit much, depends whether we expect people to know the concept. 🤔


guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
 
+## Mapping over multiple parameters
+
+There are three different ways to map over multiple parameters:
+
+- **Cross-Product**: Map over 2 or more parameters and get a mapped task instance for each possible combination of inputs.
+- **Sets of keyword arguments**: Map over 2 or more sets of one or more keyword arguments. Get a mapped task instance for every set.
+- **Zip**: Use Python's built-in `zip()` function to create sets of arguments for one parameter that is to be mapped over. Get one mapped task for every set of arguments.
+
+## Cross-Product
```
kentdanas, line 75:
```suggestion
### Cross-Product
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
 
 ![Mapped Grid](https://assets2.astronomer.io/main/guides/dynamic-tasks/mapped_grid_view.png)
 
+## Mapping over multiple parameters
+
+There are three different ways to map over multiple parameters:
+
+- **Cross-Product**: Map over 2 or more parameters and get a mapped task instance for each possible combination of inputs.
```
kentdanas, line 71:
```suggestion
- **Cross-Product**: Mapping over 2 or more parameters results in a mapped task instance for each possible combination of inputs.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- **Sets of keyword arguments**: Map over 2 or more sets of one or more keyword arguments. Get a mapped task instance for every set.
```
kentdanas, line 72:
```suggestion
- **Sets of keyword arguments**: Mapping over 2 or more sets of one or more keyword arguments results in a mapped task instance for every set.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- **Zip**: Use Python's built-in `zip()` function to create sets of arguments for one parameter that is to be mapped over. Get one mapped task for every set of arguments.
```
kentdanas, line 73:
```suggestion
- **Zip**: Mapping over a set of arguments created with Python's built-in `zip()` results in one mapped task for every set of arguments.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+## Cross-Product
+
+The default behavior of the `expand()` function is to create a mapped task instance for every possible combination of all provided inputs. For example, if you were to map over 3 keyword arguments and provide 2 options to the first, 4 options to the second and 5 options to the third, you would create 2x4x5=40 mapped task instances. One use case of this behavior is in tuning model hyperparameters.
```
kentdanas, line 77:
```suggestion
The default behavior of the `expand()` function is to create a mapped task instance for every possible combination of all provided inputs. For example, if you map over 3 keyword arguments and provide 2 options to the first, 4 options to the second, and 5 options to the third, you would create 2x4x5=40 mapped task instances. One common use case for this method is tuning model hyperparameters.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+The task definition below maps over 3 options for `bash_command` and 3 options for `env` variables. This will result in 3x3=9 mapped task instances. Each bash command runs with each definition for the environment variable `WORD`.
```
kentdanas, line 79:
```suggestion
The task definition below maps over 3 options for  the `bash_command` parameter and 3 options for the `env` parameter. This will result in 3x3=9 mapped task instances. Each bash command runs with each definition for the environment variable `WORD`.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- Map index 3: `5`
+- Map index 4: `3`
+- Map index 5: `7`
+- Map index 6: `hXllo`
+- Map index 7: `tXa`
+- Map index 8: `goodbyX`
+
+### Sets of keyword arguments
+
+To expand over sets of inputs to two or more keyword arguments (kwargs) you can use the `expand_kwargs()` function. The two tasks below show how you can provided sets of parameters as a list containing a dictionary or as an XComArg. Each operator has 3 sets of commands resulting in 3 mapped task instances.
```
kentdanas, line 112:
```suggestion
To map over sets of inputs to two or more keyword arguments (kwargs), you can use the `expand_kwargs()` function. The two tasks below show how you can provide sets of parameters as a list containing a dictionary or as an XComArg. Each operator has 3 sets of commands resulting in 3 mapped task instances.
```

kentdanas, line 112: Can we add something to clarify when you would want to use this? Overall it seems like essentially the same thing as using zip, but I think not everything can be passed as a kwarg?

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+Both `t1` and `t2` will each have 3 mapped task instances printing their results into the logs:
+
+- Map index 0: `hello`
+- Map index 1: `3`
+- Map index 2: `goodbyX`
+
+### Zip
+
+To provide sets of configurations to the same parameter, for example as arguments to a Python function, you can leverage the built-in Python function [`zip()`](https://docs.python.org/3/library/functions.html#zip).
```
kentdanas, line 146: I do think it would be helpful here to add your general zip example from your comment in the text.

guides/dynamic-tasks.md
```diff
@@ -64,6 +64,120 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+The code snippet below shows how a list of zipped arguments can be provided to the `expand()` function in order to create mapped tasks over pairs of inputs to the same argument created by the `zip()` function. It is of course possible to generate the lists of inputs with an upstream function.
+
```
kentdanas, line 149: Would zip not work with the Bash example used in the previous two sections? Seems like ideally we would use the same example in all three to help people understand the difference.

kentdanas (changes requested):
Thanks for this @TJaniF! Love the new sections and layout, I think that's perfect. Couple of comments/questions throughout, let me know if you want to chat through them next week.

guides/dynamic-tasks.md
TJaniF: I was also not fully clear on when to use what either (and I didn't realize that there is a .zip method too) ... 

I played around with it some more and the difference is that zip bundles together positional arguments in a zippy way to be provided to only one kwarg of the operator. So for example to op_args of the PythonOperator, which then gets unpacked in the python function. 
Meanwhile expand_kwargs allows to iterate in a zippy way over different sets of keyword arguments. 

I tried to make this clearer in the intro to the section and in the zip-section.

guides/dynamic-tasks.md
TJaniF: Yes, I agree. I tried to make it work but none of the BashOperator parameters takes positional arguments. 🤔 It seems when zipping all the things that get zipped have to be provided to one kwarg of the operator because .expand() does not allow any other input... 
I think the main use for zipping is really for the PythonOperator and -decorator. Unless I am grossly misunderstanding something. 👀

TJaniF (commented):
Made some changes to the multiple parameters section trying to clarify which of the 3 options does what (after clarifying it in my head 😅). 

While playing around with mapping I noticed that it isn't always super intuitive how to pass XComs from one task to be mapped over by another task especially if one of the tasks does not use the TaskFlowAPI. I decided to add a section on this, but if it makes the guide too long am of course open to deleting it. I gathered the examples in a DAG in our 2.4 dags repo, so there is the option of just linking to that DAG too. 🙂

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- Map index 3: `5`
+- Map index 4: `3`
+- Map index 5: `7`
+- Map index 6: `hXllo`
+- Map index 7: `tXa`
+- Map index 8: `goodbyX`
+
+### Sets of keyword arguments
+
+To map over sets of inputs to two or more keyword arguments (kwargs), you can use the `expand_kwargs()` function. The two tasks below show how you can provide sets of parameters as a list containing a dictionary or as an XComArg. Each operator has 3 sets of commands resulting in 3 mapped task instances.
```
kentdanas, line 205:
```suggestion
To map over sets of inputs to two or more keyword arguments (kwargs), you can use the `expand_kwargs()` function. The two tasks below show how you can provide sets of parameters as a list containing a dictionary or as an `XComArg`. Each operator has 3 sets of commands, resulting in 3 mapped task instances.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+```Python
+# input sets of kwargs directly as a list[dict] PENDING IMPLEMENTATION!
```
kentdanas, line 208: Just a reminder to remove the PENDING IMPLEMENTATION note when ready - I would totally forget to do this :)

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+t1 = BashOperator.partial(task_id="t1").expand_kwargs(
+    [
+        {"bash_command": "echo $WORD", "env" : {"WORD": "hello"}},
+        {"bash_command": "echo `expr length $WORD`", "env" : {"WORD": "tea"}},
+        {"bash_command": "echo ${WORD//e/X}", "env" : {"WORD": "goodbye"}}
+    ]
+)
+
+# input sets of kwargs generated by another task as an XComArg object
```
kentdanas, line 217: Now that you've added a section on mapping on the results of another task, do you think we still need this t2 here? If not I think the example would be more straight forward with a single task...

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+- `zip(["a", "b", "c"], [1, 2, 3], ["hi", "bye", "tea"])` will result in a zip object containing: `("a", 1, "hi"), ("b", 2, "bye"), ("c", 3, "tea")`
+- `zip(["a", "b"], [1], ["hi", "bye"], [19, 23], ["x", "y", "z"])` will result in a zip object containing only one tuple: `("a", 1, "hi", 19, "x")` because the shortest list provided only contains one element.
+- It is also possible to zip together different types of iterables: `zip(["a", "b"], {"hi", "bye"}, (19, 23))` will result in a zip object containing: `('a', 'hi', 19), ('b', 'bye', 23)`.
+
+> **Note**: To get as many tuples as there are elements in the *longest* list provided you can use the [`zip_longest()`](https://docs.python.org/3/library/itertools.html#itertools.zip_longest) function from the Python package `itertools`.
+
+The code snippet below shows how a list of zipped arguments can be provided to the `expand()` function in order to create mapped tasks over sets of positional arguments. Each set of positional arguments is passed to a keyword argument, `zipped_x_y_z` in case of the TaskFlow example and `op_args` in case of the `PythonOperator` example.
```
kentdanas, line 249: Let's update this example to be only TaskFlow, and decorate the `add_numbers` function. Astronomer is recommending taskflow over regular PythonOperator now, and zip is basically the same for both so I think we can simplify.

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+The `zip()` function takes in an arbitrary number of iterables (for example lists) and uses their elements to create a zip-object containing tuples. There will be as many tuples as there are elements in the shortest iterable. Each tuple contains one element from every iterable provided. For example:
+
+- `zip(["a", "b", "c"], [1, 2, 3], ["hi", "bye", "tea"])` will result in a zip object containing: `("a", 1, "hi"), ("b", 2, "bye"), ("c", 3, "tea")`
+- `zip(["a", "b"], [1], ["hi", "bye"], [19, 23], ["x", "y", "z"])` will result in a zip object containing only one tuple: `("a", 1, "hi", 19, "x")` because the shortest list provided only contains one element.
+- It is also possible to zip together different types of iterables: `zip(["a", "b"], {"hi", "bye"}, (19, 23))` will result in a zip object containing: `('a', 'hi', 19), ('b', 'bye', 23)`.
+
+> **Note**: To get as many tuples as there are elements in the *longest* list provided you can use the [`zip_longest()`](https://docs.python.org/3/library/itertools.html#itertools.zip_longest) function from the Python package `itertools`.
```
kentdanas, line 247: I take it from the text below that you don't actually use `zip_longest` in the context of task mapping (instead you would use `fillvalue`)? If that's the case I would remove this note.

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+add_nums.expand(zipped_x_y_z=zipped_arguments)
+```
+
+The add_nums task will have three mapped instances with the following results:
+
+- Map index 0: `111` (1+10+100)
+- Map index 1: `1202` (2+1000+200)
+- Map index 2: `2300` (1000+1000+300)
+
+> **Note**: Mixing direct inputs of iterables with inputs via XComs in the `zip()` function is currently not supported.
```
kentdanas, line 327: I find this a bit confusing as it's currently worded - maybe something like "Zipping inputs defined in your DAG script with inputs pulled from XCom is not currently supported"? If I'm reading it right...

guides/dynamic-tasks.md
```diff
@@ -64,6 +65,267 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+).expand(op_args=zipped_arguments)
+```
+
+Both tasks, `TaskFlow_add_numbers` and `standard_add_numbers` will have three mapped task instances each performing one calculation with the results being:
+
+- Map index 0: `111`
+- Map index 1: `222`
+- Map index 2: `333`
+
+It is also possible to 'zip' XComArg objects by using their `.zip()` method to combine them with one or more other XComArg objects. If the upstream task has been defined using the TaskFlowAPI simply provide the function call. If the upstream task used a classical operator, provide the `XComArg(task_object)`. Below you can see an example of the results of two TaskFlowAPI tasks and one classical operator being zipped together to form functionally the same `zipped_arguments` (`[(1,10,100), (2,20,200), (3,30,300)]`) as in the previous example.
```
kentdanas, line 282:
```suggestion
It is also possible to zip `XComArg` objects. If the upstream task has been defined using the TaskFlow API, simply provide the function call. If the upstream task used a traditional operator, provide the `XComArg(task_object)`. Below you can see an example of the results of two TaskFlowAPI tasks and one traditional operator being zipped together to form the same `zipped_arguments` (`[(1,10,100), (2,20,200), (3,30,300)]`) as in the previous example.
```

kentdanas (commented):
@TJaniF this looks great! Agree with adding the additional section, I think it's helpful. I made some comments throughout on potentially simplifying the examples, but after that let's move on to copy editing.

guides/dynamic-tasks.md
TJaniF: What I was trying to say is that both 

```Python 
zipped_arguments = list(zip(one_two_three(), [10,20,30))
``` 

does not work (the error is " 'XComArg' object is not iterable"); 

and

```Python
zipped_arguments = one_two_three().zip(
        [10,20,30],
        XComArg(one_two_three_hundred_task),
        fillvalue=1000
    )
```

doesn't work. It fails with the error "'list' object has no attribute 'iter_references'". That is what I meant with "no mixing" 😅 

Yeah it is confusing. I think I actually need to make the distinction clearer when introducing zip() and .zip()... I ended up putting the two in different subsections and contrasting them at the start of the "zip" section. I hope this makes more sense. :)

TJaniF (commented):
Thank you for the review! I think the examples are much easier to follow now. I also divided the ### Zip section into two subsections one for the Python builtin function, one for the XComArg method. I hope this makes the distinction clearer :) 

@jwitz I think the update is ready for you 🙂

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+def one_two_three_TF():
+    return [1,2,3]
+
+@task
+def plus_10_TF(x):
+    return x+10
+
+plus_10_TF.partial().expand(x=one_two_three_TF())
+```
+
```
jwitz, line 94:
```suggestion
### Map inputs to an operator-defined task from a TaskFlow API-defined task.

```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+Passing data from an upstream task defined using the TaskFlowAPI to a downstream traditional operator works in a very similar fashion. Note that the format of the mapping information returned by the upstream task may need to be modified to be accepted by the `op_args` argument of the traditional `PythonOperator`.
```
jwitz, line 95:
```suggestion
Passing data from an upstream task defined using the TaskFlowAPI to a downstream traditional operator works in a very similar fashion. Note that the format of the mapping information returned by the upstream TaskFlow API task might need to be modified to be accepted by the `op_args` argument of the traditional PythonOperator.
```

jwitz: Is the last sentence specific to the PythonOperator or does that advice apply to all traditional operators?

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+    return x+10
+
+plus_10_task = PythonOperator.partial(
+    task_id="plus_10_task",
+    python_callable=plus_10_traditional
+).expand(
+    op_args=one_two_three_TF()
+)
+```
+
```
jwitz, line 113:
```suggestion

### Map inputs to  TaskFlow API-defined task from an operator-defined task.

```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+If you are mapping over the results of a traditional operator, you need to extract the return value using the `XComArg` object.
```
jwitz, line 114:
```suggestion
If you are mapping over the results of a traditional operator, you need to format the argument for `expand()` using the `XComArg` object.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+    return x+10
+
+one_two_three_task = PythonOperator(
+    task_id="one_two_three_task",
+    python_callable=one_two_three_traditional
+)
+
+plus_10_TF.partial().expand(x=XComArg(one_two_three_task))
+```
+
```
jwitz, line 133:
```suggestion

### Map inputs when both tasks are defined with traditional operators


```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+# when only using traditional operators, define dependencies explicitly
+one_two_three_task_2 >> plus_10_task_2
+```
+
+## Mapping over multiple parameters
+
+There are three different ways to map over multiple parameters:
+
+- **Cross-Product**: Mapping over 2 or more *keyword* arguments results in a mapped task instance for each possible combination of inputs. This type of mapping uses the `expand()` function.
```
jwitz, line 166:
```suggestion
- **Cross-product**: Mapping over 2 or more *keyword* arguments results in a mapped task instance for each possible combination of inputs. This type of mapping uses the `expand()` function.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- **Sets of keyword arguments**: Mapping over 2 or more sets of one or more *keyword* arguments results in a mapped task instance for every set. This type of mapping uses the `expand_kwargs()` function.
```
jwitz, line 167:
```suggestion
- **Sets of keyword arguments**: Mapping over 2 or more sets of one or more keyword arguments results in a mapped task instance for every defined set, rather than every combination of individual inputs. This type of mapping uses the `expand_kwargs()` function.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+    ],
+    env=[
+        {"WORD": "hello"},
+        {"WORD": "tea"},
+        {"WORD": "goodbye"}
+    ]
+)
+```
+
+The nine mapped task instance of the task `cross_product_example` run all possible combinations of the bash command with the env variable:
```
jwitz, line 193:
```suggestion
The nine mapped task instances of the task `cross_product_example` run all possible combinations of the bash command with the `env` variable:
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+The task `t1` will have 3 mapped task instances printing their results into the logs:
+
+- Map index 0: `hello`
+- Map index 1: `3`
+- Map index 2: `goodbyX`
+
+### Zip
+
+In Airflow 2.4+ you can provide sets of positional arguments to the same keyword argument, for example to the `op_args` keyword argument of the `PythonOperator`. You can use the built-in Python function [`zip()`](https://docs.python.org/3/library/functions.html#zip) if your inputs are in the form of iterables such as tuples, dictionaries or lists and the `.zip()` method of the `XComArg` object if your inputs are coming from XComs.
```
jwitz, line 228:
```suggestion
In Airflow 2.4+ you can provide sets of positional arguments to the same keyword argument, such as the `op_args` argument of the PythonOperator. You can use the built-in [`zip()`] function(https://docs.python.org/3/library/functions.html#zip) Python function if your inputs are in the form of iterables such as tuples, dictionaries, or lists. If your inputs come from XCom objects, you can use the `.zip()` method of the `XComArg` object.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+> **Note**: It is currently not possible to provide XComArg objects to the built-in Python function `zip()` or to directly provide iterables to the `.zip()` method of the XComArg object.
```
jwitz, line 230:
```suggestion
```

I think this is evident from the text above

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+#### Built-in Python function zip()
+
+The `zip()` function takes in an arbitrary number of iterables (for example lists) and uses their elements to create a zip-object containing tuples. There will be as many tuples as there are elements in the shortest iterable. Each tuple contains one element from every iterable provided. For example:
+
+- `zip(["a", "b", "c"], [1, 2, 3], ["hi", "bye", "tea"])` will result in a zip object containing: `("a", 1, "hi"), ("b", 2, "bye"), ("c", 3, "tea")`
```
jwitz, line 236:
```suggestion
- `zip(["a", "b", "c"], [1, 2, 3], ["hi", "bye", "tea"])` results in a zip object containing: `("a", 1, "hi"), ("b", 2, "bye"), ("c", 3, "tea")`.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- `zip(["a", "b"], [1], ["hi", "bye"], [19, 23], ["x", "y", "z"])` will result in a zip object containing only one tuple: `("a", 1, "hi", 19, "x")` because the shortest list provided only contains one element.
```
jwitz, line 237:
```suggestion
- `zip(["a", "b"], [1], ["hi", "bye"], [19, 23], ["x", "y", "z"])` results in a zip object containing only one tuple: `("a", 1, "hi", 19, "x")`. This is because the shortest list provided only contains one element.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- It is also possible to zip together different types of iterables: `zip(["a", "b"], {"hi", "bye"}, (19, 23))` will result in a zip object containing: `('a', 'hi', 19), ('b', 'bye', 23)`.
```
jwitz, line 238:
```suggestion
- It is also possible to zip together different types of iterables. `zip(["a", "b"], {"hi", "bye"}, (19, 23))` results in a zip object containing: `('a', 'hi', 19), ('b', 'bye', 23)`.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+The code snippet below shows how a list of zipped arguments can be provided to the `expand()` function in order to create mapped tasks over sets of positional arguments. Each set of positional arguments is passed the keyword argument `zipped_x_y_z`.
```
jwitz, line 240:
```suggestion
The code snippet below shows how a list of zipped arguments can be provided to the `expand()` function in order to create mapped tasks over sets of positional arguments. Each set of positional arguments is passed in the keyword argument `zipped_x_y_z`.
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+# creating the mapped task instances using the TaskFlowAPI
+@task
+def add_numbers(zipped_x_y_z):
+    return zipped_x_y_z[0] + zipped_x_y_z[1] + zipped_x_y_z[2]
+
+add_numbers.expand(zipped_x_y_z=zipped_arguments)
+```
+
+The task `add_numbers` will have three mapped task instances each performing one calculation with the results being:
```
jwitz, line 255:
```suggestion
The task `add_numbers` will have three mapped task instances for each possible combination of positional arguments:
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+- Map index 1: `3`
+- Map index 2: `goodbyX`
+
+### Zip
+
+In Airflow 2.4+ you can provide sets of positional arguments to the same keyword argument, for example to the `op_args` keyword argument of the `PythonOperator`. You can use the built-in Python function [`zip()`](https://docs.python.org/3/library/functions.html#zip) if your inputs are in the form of iterables such as tuples, dictionaries or lists and the `.zip()` method of the `XComArg` object if your inputs are coming from XComs.
+
+> **Note**: It is currently not possible to provide XComArg objects to the built-in Python function `zip()` or to directly provide iterables to the `.zip()` method of the XComArg object.
+
+#### Built-in Python function zip()
```
jwitz, line 232:
```suggestion
#### Provide positional arguments with the built-in Python zip()
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+add_numbers.expand(zipped_x_y_z=zipped_arguments)
+```
+
+The task `add_numbers` will have three mapped task instances each performing one calculation with the results being:
+
+- Map index 0: `111`
+- Map index 1: `222`
+- Map index 2: `333`
+
+#### .zip() method of the XComArg object
```
jwitz, line 261:
```suggestion
#### Provide positional arguments XComArg object zip()
```

guides/dynamic-tasks.md
```diff
@@ -64,6 +67,244 @@ Similarly, the Grid View shows task details and history for each mapped task. Al
...
+
+It is also possible to zip `XComArg` objects. If the upstream task has been defined using the TaskFlow API, simply provide the function call. If the upstream task used a traditional operator, provide the `XComArg(task_object)`. Below you can see an example of the results of two TaskFlowAPI tasks and one traditional operator being zipped together to form the `zipped_arguments` (`[(1,10,100), (2,1000,200), (1000,1000,300)]`).
+
+To mimic the behavior of the [`zip_longest()`](https://docs.python.org/3/library/itertools.html#itertools.zip_longest) function from the `itertools` package, you can add an optional keyword argument `fillvalue` to the `.zip()` method. If `fillvalue` is specified, the method will produce as many tuples as the longest input has elements, missing elements are filled with the default value. If `fillvalue` would not have been specified in the example below, `zipped_arguments` would only contain one tuple `[(1,10,100)]` since the shortest list provided to the `.zip()` method is only one element long.
```
jwitz, line 265:
```suggestion
To mimic the behavior of the [`zip_longest()`](https://docs.python.org/3/library/itertools.html#itertools.zip_longest) function, you can add the optional `fillvalue` keyword argument to the `.zip()` method. If you specify a default value with `fillvalue`, the method produces as many tuples as the longest input has elements and fills in missing elements with the default value. If `fillvalue` was not specified in the example below, `zipped_arguments` would only contain one tuple `[(1,10,100)]` since the shortest list provided to the `.zip()` method is only one element long.
```

jwitz (approved):
Looks good! Nice job on breaking down some very complex concepts and implementations. I tried making the guide slightly easier to read by adding some more headers, but feel free to play around with my wording there and in other suggestions!

kentdanas (approved):

TJaniF:
DO NOT MERGE until conforming that the example on line 216 has been implemented and take out the warning comment. :) (other than that proofreading is done and ready to merge)

TJaniF (commented):
@jwitz I have no idea how I didn't see your review the last time I read through a week ago 🤔 . Sorry about that and thank you! 

Other than your edits I took out the "pending implementation" and switched out a screenshot to the 2.4 version (showing mapped task instances in the grid view). Ready to merge!

guides/dynamic-tasks.md
```diff
@@ -7,42 +7,51 @@ heroImagePath: null
...
 
-With the release of [Airflow 2.3](https://airflow.apache.org/blog/airflow-2.3.0/), users can write DAGs that dynamically generate parallel tasks at runtime. This feature, known as dynamic task mapping, is a paradigm shift for DAG design in Airflow. 
-
-Prior to Airflow 2.3, tasks could only be generated dynamically at the time that the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. With dynamic task mapping, you can easily write DAGs that create tasks based on your current runtime environment. 
+Prior to Airflow 2.3, tasks could only be generated dynamically at the time that the DAG was parsed, meaning you had to change your DAG code if you needed to adjust tasks based on some external factor. With dynamic task mapping, you can easily write DAGs that create tasks based on your current runtime environment.
 
 In this guide, we'll explain the concept of dynamic task mapping and provide example implementations for common use cases.
```
TJaniF, line 14:
```suggestion
In this guide, we'll explain the concept of dynamic task mapping and provide an example implementation for a common use case.
```

PR #433, opened 2022-09-20
Fix syntax highlighting - dynamic tasks guide

TJaniF:


kentdanas:
Thanks @TJaniF! I think we also need to update in the scheduling guide and the cross-dag dependencies guide

kentdanas (approved):

